PR #1, opened 2015-02-27
Updating gulp setup and adding UMD wrapping around modules.

mandragorn:
Some initial work, let me know what you think and I can continue.


kentmw:
Is looking good. Only question is do we have to depend on all of torso's dependencies on the sub modules like form model or just torso

arikwex:
Why are we all awake....?

mandragorn:
heh, went to sleep immediately after pushing.

@kentmw  - that was a question I was going to ask in the morning.  We can do it 2 ways.
1. Assume that if you aren't using globals then you will always use some AMD type spec and each object lives detached.
2. Maintain the Torso object and attach stuff to it, but don't expose it as a global.

I don't think 2 will work since it will be harder to guarantee that the aspect of Torso you want to use has already been loaded (plus it is weird when using that paradigm).  So, in the case of non-global imports you import each piece separately (and only use the Torso dependency if you need something from the core of torso... which might beg the question of whether that stuff should be exposed separately).  In the case of global imports its all tied together anyway.

mandragorn:
Not sure if I did everything right when updating the request and do you know if the merging of the pull request will result in one commit or 3?  Is there a way to collapse my commits into a single commit when its merged to clean up the history?

PR #2, opened 2015-03-02
Updating gulp setup and adding UMD wrapping around modules.

mandragorn:
Sorry I did some work to clean up the intermediate commits to prove to myself that its possible, so this is a new pull request.

Still needs cleanup and discussion, key point is that this passes unit tests.

What things do I need to check that are not under unit test?


mandragorn:
This is a bit of a mess right now as I'm trying to figure out how to get commonjs tests to work (requiring directly in gulp didn't work because backbone models required a dom to be present in the environment).

mandragorn:
Majority of commonJs tests implemented and modules upgraded to UMD.  Still a few to flesh out and get the original tests working again.

mandragorn:
I think this is ready to go, at least ready for a full review.  Everything works and the gulp jobs are cleaned up, etc.

mandragorn:
I made some significant changes to the module layout under Torso, let me know what you think.

mandragorn:
I think I'm done making changes now... pending peer review comments.

mandragorn:
note: don't merge this pull request I'm planning on squashing it into a single commit on another branch after addressing peer review comments.

PR #9, opened 2015-03-09
Cleaned up script locations (/modules) and gulp configurations.

mandragorn:
moved code into flat modules directory for easy import of parts into other projects.
Simplified and streamlined gulp build
For gulp watch and gulp tests:watch to work gulp-jasmine needs to be updated with these changes: https://github.com/sindresorhus/gulp-jasmine/pull/43


mandragorn:
bloody hell that was annoying.... look at the overall diff and once I address peer review comments I'll merge this into a single commit and push a new pull request.

README.md
```diff
@@ -2,116 +2,82 @@
 A holistic approach to Backbone applications
 
 ## Build Commands:
```
kentmw: Let's chat about how we want to set up our readme. I think this is too much about the build.

README.md
```diff
@@ -2,116 +2,82 @@
...
-npm run-script build
-npm run-script build:clean
+npm run build
 ```
 or
 ```
-node_modules/gulp/bin/gulp.js default
-node_modules/gulp/bin/gulp.js default:clean
-
-node_modules/gulp/bin/gulp.js build
```
kentmw: I think any references to gulp should be 'gulp' not 'node_modules/...gulp'

modules/torso.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+    events: events,
+    Mixins: {
+      collectionLoading: collectionLoadingMixin,
+      collectionRegistration: collectionRegistrationMixin,
+      polling: pollingMixin,
+      validation: validation.mixin,
+      viewHierarchy: viewHierarchyMixin
+    },
+    Models: {
```
kentmw: this matches what is generated when requiring all the modules globally, correct?
Not a big fan of Torso.Models.Model, but we can think about that later.

test/spec/import/commonJsImports.js
```diff
@@ -0,0 +1,81 @@
+var _ = require('underscore'),
+    commonJsImportTest = function(moduleToImport, expectedModules) {
```
kentmw: define the function variable name before _ and then define the function without a tab

test/spec/import/globalImports.js
```diff
@@ -0,0 +1,49 @@
+describe('Global Module imports', function() {
+  var window, globalIndex,
+      globals = [
+        ['Torso'],
```
kentmw: Add a comment about what this set up does

test/spec/import/commonJsImports.js
```diff
@@ -0,0 +1,81 @@
...
+      expectedModules.push(moduleToImport);
+      describe('CommonJS Module import of "' + moduleToImport + '"', function() {
+        var windowRequire, module, actualModule, moduleIndex;
+
+        beforeAll(function(done) {
```
kentmw: give an example of what this does in comments

modules/torso.js
mandragorn: Yea but before it was:
Torso.Model
Torso.Models.Nested
which seemed inconsistent to me.  At least this is consistent for all models.

PR #13, opened 2015-04-15
List view updates and adding list view tests and #20 rename torso bundle hierarchy

kentmw:


modules/ListView.js
```diff
@@ -22,7 +21,44 @@
...
+     */
+    aggregateRenders = function(wait, view) {
+      var postpone = function() {
+        view._delayedRenderTimeout = null;
+        view.render();
+      };
+      return function() {
+        if (!view._delayedRenderTimeout && wait > 0) {
+          view._delayedRenderTimeout = setTimeout(postpone, wait);
+        } else if (wait === 0) {
```
mandragorn: do we care about <= 0?

kentmw: I don't know what that would mean. Or are you just doing that to be safe

mandragorn: yea, just to validate the input before using it.  I could see using wait = -1 for no wait.

modules/ListView.js
```diff
@@ -22,7 +21,44 @@
    */
   var ListView = View.extend((function() {
 
-    var removeChildView, addChildView;
+    var removeChildView, addChildView, aggregateRenders, breakDelayedRender;
```
mandragorn: Why the additional functional scoping?  Aren't multiple ListViews going to get the same context anyway?  It would be clearer what is 'super private' if it lived outside the View.extend()

kentmw: done

modules/ListView.js
kentmw: done

kentmw:
Any other comments?

Issue #30, opened 2015-04-27
backbone-stickit needs to move to dependencies from devDependencies


kentmw:
When being brought in to a new project, we also have to bring in backbone-stickit because Torso needs it, but it's not registered as a dependency.

Issue #29, opened 2015-04-27
ListView uses $ with no factory input


arikwex:
Hey.
List View crashes when $ isn't in public scope.
List View uses $ instead of this.$ inside the class.
Either pass a jquery instance in to the factory OR modify to us this.$

kentmw:
Added jquery as a dependency in the module definition

Issue #27, opened 2015-04-27
torso-bundle.js is not version controlled


kentmw:

kentmw:
I think I understand why we have a bundle - for when someone want to just grab the bundle for global use. However, we .gitignore the bundle, which means someone has to pull the project and build the code in order to get it. I vote we either version control the bundle or move it to test

mandragorn:
torso-bundle.js should be included when it is published to NPM.  So version controlling it makes sense as long as we have some way to make sure it is kept up to date since the build process will change it based on other files.

Issue #14, opened 2015-04-26
Separate npm install and npm prepublish


kentmw:
Right now npm install builds the code needed to publish. However, npm install is run on projects to get the post-published code. Therefore, we need to separate the concept of "npm install"-ing the code (which will be run for everyone using the code) and building the code before publishing - which happens while dev'ing on Torso

mandragorn:
I'm not sure I understand, there currently isn't any pre-publish or install scripts setup in the package.json.

kentmw:
This is true. I don't mean separating actual scripts - just that we need to separate what's functionally happening when you run npm install.

kentmw:
I can't replicate this issue. Maybe I was wrong, or maybe the npm prepublish fixed it.

mandragorn:
this may have been an old issue while I was working things out in the build process.  At one point I had install tied to a script in package.json.

PR #33, opened 2015-04-29
#30, #29, #27, #14 and adding view state cell

kentmw:
Fixing some of the publishing/prepublishing issues, added view state cell, and fixed list view's jquery dependency.


modules/View.js
```diff
@@ -118,9 +120,10 @@
       this.remove();
 
       // Unbind all local event bindings
-      this.unbind();
       this.off();
       this.stopListening();
+      this.viewState.off();
```
kentmw: check if view state exists

kentmw:
Ariel Wexler gave a +2

PR #46, opened 2015-05-07
Fix templateRenderer not rendering text node changes

andyhuang91:
Update TemplateRenderer.hotswap to correctly process text and comment nodes
Refactor TemplateRenderer.hotswap
Add unit tests for TemplateRenderer.hotswap


torso-bundle.js
```diff
@@ -256,149 +369,61 @@
...
-      return hardRefreshes;
-    },
+    // See above function declaration for method-level documentation
+    hotswap: hotswap,
 
     /**
      * Produces a copy of the element tag with attributes but with no contents
-     * @param el {jQuery element} the element to be copied
-     * @return a shallow copy of the element with no children but with attributes
+     * @param el {Element} the DOM element to be copied
```
andyhuang91: I'm changing the interface for the copyTopElement method to take and return plain Element objects instead of jQuery objects. This is why I made a same change to ListView

PR #69, opened 2015-06-18
Added automatic generation of select options for stickit bindings.

ccpowers:


modules/FormView.js
```diff
@@ -382,6 +387,29 @@
     },
 
     /**
+     * @method  _generateSelectOptions
+     * @param element {Element} The select element to generate options for
+     * @param opts {Object} Additional behavior options for the bindings
+     * @param [options.modelFormat] {Object} The function called before setting model values
```
kentmw: opts.modelFormat?

modules/FormView.js
```diff
@@ -382,6 +387,29 @@
     },
 
     /**
+     * @method  _generateSelectOptions
```
kentmw: double underscore: __generateSelectOptions

modules/FormView.js
```diff
@@ -382,6 +387,29 @@
...
+    _generateSelectOptions: function(element, opts) {
+      var collection = [],
+          options = $(element).children('option');
+
+      _.each(options, function(option) {
+        collection.push({'label': $(option).text(),
+                          'value': opts.modelFormat ? opts.modelFormat.apply(this, $(option).val()) : $(option).val()});
+      });
+
+      return {collection: collection,
```
kentmw: Add an issue to Torso that explains the bug and reference the issue in this commit message.

modules/FormView.js
ccpowers: Done.

modules/FormView.js
ccpowers: Done.

modules/FormView.js
ccpowers: Done. See #84.

modules/FormView.js
```diff
@@ -382,6 +387,29 @@
...
+     * @param [opts.modelFormat] {Object} The function called before setting model values
+     * @private
+     * @return {<Stickit select options hash>}
+     */
+    __generateSelectOptions: function(element, opts) {
+      var collection = [],
+          options = $(element).children('option');
+
+      _.each(options, function(option) {
+        collection.push({'label': $(option).text(),
```
kentmw: last question. This all looks good, but I'm a little worried about blowing away your select options if they have different classes/attributes that you had added before. Will you lose classes?

ccpowers: I just tested this- you will lose classes on the select options. You won't lose data on the actual select though. There might be a clever way to get around this with the 'classes' bindings available in stickit - I will think about it some more.

ccpowers: There doesn't seem to be a way to keep your classes while binding select options using selectOptions - any ideas?

kentmw: @ccpowers, I think you should merge this. Though, let's create an enhancement to keep classes on two-way bound select options.

kentmw: @ccpowers, do you want to keep working on this or should I take over?

ccpowers: @kentmw I'm happy to finish it but i'm about to leave the country for a month so if you want it done before October, you should probably take over...just let me know

kentmw: We can wait until you get back, that will be fine. If we need it before you return, we'll create a new PR. Have fun!

kentmw: Hey @ccpowers, are you back? Any interest in picking this back up? We're pretty close to finishing.

PR #89, opened 2015-07-09
Added the ability for ListView Children to use different ChildViews

patrickckelly:


kentmw:
handled in another pull request

PR #92, opened 2015-07-10
Added the ability for ListView Children to use different ChildViews

patrickckelly:
This works in 0.1.x


arikwex:
Looks good.  Need to merge against the most recent torso 0.1.x and submit again.  Also, this patch should get pull-requested for master as well when ready.

kentmw:
Handled in another pull request

PR #98, opened 2015-08-04
add appveyor config file

catbieber:
Appveyor will build on Windows. Torso currently fails to build on Windows. Hoping this can increase visibility and, once we get the build passing, keep it passing.


kentmw:
I think @arikwex had issues with building on windows. Issues with jsdom, I believe

catbieber:
Yup, not sure if this is the same failure or not...apparently I need to install Visual Studio to be able to build torso :(. I wasn't able to get even as far as Appveyor does, to confirm whether it's the same issue.

PR #99, opened 2015-08-04
Adding Logger module to track events

cmchin:
Questions: 
- what if theres a start signal but no end signal? How long should we wait before we assume that its the end of the session and something went wrong? 
- How much can I assume that evenInfo can has all the keys that I want it to? (For example, how do I handle events that don't have a uuid or session id?) 
- Should I not send information to new relic in logger if we're still not sure if we're going to buy it?
- How do I minimize the amount of code I copy and paste from backbone/torso?


modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+    _sessionStart: null,
+    _sessionEnd: null,
+    _sessionID: null,
+
+    /**
+    * saves information about event and wraps start and end signals in 'session' 
+    * @ param {JSON object} contains tracking information (including state: 'start' or 'end', uuid, sessionID)
+    * @ method track
+    */
+  	track: function(eventInfo){
```
kentmw: indentation issue

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+    sessions: [],
+    unclosed: {},
+    _isClosed: true,
+    _sessionStart: null,
+    _sessionEnd: null,
+    _sessionID: null,
+
+    /**
+    * saves information about event and wraps start and end signals in 'session' 
+    * @ param {JSON object} contains tracking information (including state: 'start' or 'end', uuid, sessionID)
```
kentmw: @param and @method not @ param and @ method
Also, usually, @param is followed by the name of the parameter and then the type

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+  } else if (typeof exports === 'object') {
+    module.exports = factory(require('underscore'), require('backbone'), require('./ServiceCell'));
+  } else {
+    root.Torso = root.Torso || {};
+    root.Torso.Logger = factory(root._, root.Backbone, root.Torso.ServiceCell);
+  }
+}(this, function(_, Backbone, ServiceCell) {
+  'use strict';
+
+  var Logger = ServiceCell.extend({
```
kentmw: I wonder if Logger is the right name for this service. Logger typically concerns logging info/debug/warn/error. This sounds like a session tracker or an event tracker

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+
+    /**
+    * tracks information about the 'start' of events
+    * @param {JSON object} contains tracking information including UUID and time
+    * @method startEvent
+    */
+    startEvent: function(eventInfo){
+      var uuid = eventInfo.UUID;
+      eventInfo.sessionID = this._sessionID; //check if event info has sessionID property and throw error
+      eventInfo.startTime = eventInfo.time;
```
kentmw: for simplicity sake, can we have the event start time be generated here. Maybe something like:

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+      }
+  	},
+
+    /**
+    * tracks information about the 'start' of events
+    * @param {JSON object} contains tracking information including UUID and time
+    * @method startEvent
+    */
+    startEvent: function(eventInfo){
+      var uuid = eventInfo.UUID;
```
kentmw: Also, probably best to generate a UUID for people here, and return the UUID at the end of this method.

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+    * tracks information about the 'end' of events
+    * @param {JSON object} contains information about the tracked event, including UUID, time, state
+    * @method endEvent
+    */
+    endEvent: function(eventInfo){
+      var uuid = eventInfo.UUID;
+      var totalEvent = this.unclosed[uuid];
+      var duration = eventInfo.time - totalEvent.startTime; 
+      totalEvent.duration = duration;
+      totalEvent.endTime = eventInfo.time;
```
kentmw: Same with endTime. We probably can default the end time to when someone calls endEvent

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+    */
+    openSession: function(eventInfo){
+      this._isClosed = false;
+      var sessionID = (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+      this._sessionID = sessionID;
+      this._sessionStart = eventInfo.time;
+    },
+
+    /**
+    * tracks information about the end of a session. resets all relevant parameters
```
kentmw: Is this called manually? A little more detail in this method would be nice. Would a failure to connect to the server trigger an end to session? I guess this points out, what session are we talking about.

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+    /**
+    * send all information from the session to new relic
+    * @method sendToNewRelic
+    */
+    sendToNewRelic: function(){
+      var sessionLength = this.sessions.length;
+      for (var i = 0; i < sessionLength; i++){
+        var evt = this.sessions[i];
+        var actionName = evt.type;
+        newrelic.addPageAction(actionName, evt);
```
kentmw: there is a hidden dependency in this module on a variable "newrelic" in global scope. This needs to be explicit dependency and also optional for use with torso. We should be checking if window.newrelic exists and throwing an exception in this method if it doesn't.

modules/Logger.js
```diff
@@ -0,0 +1,169 @@
...
+      var sessionDuration = this._sessionEnd - this._sessionStart;
+      sessionEvent.sessionDuration = sessionDuration;
+      this.sessions.push(sessionEvent);
+      for (var key in this.log){
+        this.log[key].sessionDuration = sessionDuration;
+        this.sessions.push(this.log[key]);
+        delete this.log[key];
+      }
+      this._isClosed = true;
+      this.sendToNewRelic();
```
kentmw: sending to new relic should be an option when setting up this service, not done by default

modules/Model.js
```diff
@@ -1,23 +1,58 @@
...
+
+    /** 
+    * Overridden to send before and after signals to Logger
+    */
+    fetch: function(options){
+      var UUID = "uuid-"+(new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+      Logger.track({
+        UUID : UUID,
+        type : "fetch",
+        state: "start",
```
kentmw: can a new event being tracked, default to start state?

modules/Model.js
```diff
@@ -1,23 +1,58 @@
...
-   * @module    Torso
-   * @class     Model
-   * @constructor
-   * @author kent.willis@vecna.com
-   */
-  var Model = Backbone.Model.extend({});
+
+  var Model = Backbone.Model.extend({
+
+    wrapError: function(options){
```
kentmw: what does this method do? It's name isn't indicative (enough) of what it's doing

modules/Model.js
```diff
@@ -1,23 +1,58 @@
...
+      var UUID = "uuid-"+(new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+      Logger.track({
+        UUID : UUID,
+        type : "fetch",
+        state: "start",
+        time: Date.now(),
+      });
+
+      options = options ? _.clone(options) : {};
+      if (options.parse === void 0) options.parse = true;
```
kentmw: did you have to copy the original fetch method? We want to prevent this if at all possible
Can you instead call the parent's prototype and wrap that with what you want?

modules/Router.js
```diff
@@ -15,5 +15,48 @@
...
    */
-  return Backbone.Router.extend({});
+
+  var Router = Backbone.Router.extend({
+
+    /**
+    * overridden the route function to send start and end times to Logger
+    */
+    route: function(route, name, callback) {
+      if (!_.isRegExp(route)) route = this._routeToRegExp(route);
```
kentmw: Same comment here as with Model's fetch. Can we wrap the parent's prototype?

modules/View.js
```diff
@@ -56,6 +56,70 @@
...
+
+    /**
+    * updates Backbone's delegateEvents to bind signals to Logger to all callback 
+    * functions in the View's list of events
+    * @method updateDelegateEvents
+    */
+    updateDelegateEvents: function(){
+      var backboneDelegateEvents = Backbone.View.prototype.delegateEvents;
+      Backbone.View.prototype.delegateEvents = function(events){
+        var delegateEventSplitter = /^(\S+)\s*(.*)$/;
```
kentmw: I really don't want us copying the internals of backbone into torso

modules/View.js
```diff
@@ -56,6 +56,70 @@
...
+    },
+
+
+    /**
+    * @param method {function} callback method of event
+    * @param eventName {String} description of event type 
+    * @return modified mehtod with before/after signals sent to Logger
+    * @method trackEvents
+    */
+    trackEvents : function(method, eventName){
```
kentmw: is this wrapping any view method with an event start and stop?

modules/View.js
```diff
@@ -99,8 +163,30 @@
     templateRender: function(el, template, context, opts) {
       // Detach just this view's child views for a more effective hotswap.
       // The child views will be reattached by the render method.
+      var pageName = null;
+      if(opts){
+        pageName = opts.pageName;
+      }
+
+      newrelic.setCustomAttribute('pageName', pageName);
```
kentmw: newrelic dependency needs to be optional

kentmw:
To answer your questions:

modules/Model.js
```diff
@@ -1,23 +1,58 @@
...
+      options.error = function(resp) {
+        if (error) error(model, resp, options);
+        model.trigger('error', model, resp, options);
+      };
+    },
+
+    /** 
+    * Overridden to send before and after signals to Logger
+    */
+    fetch: function(options){
```
kentmw: look into $.ajaxSetup's beforeSend and complete

modules/Router.js
```diff
@@ -15,5 +15,48 @@
...
+      if (_.isFunction(name)) {
+        callback = name;
+        name = '';
+      }
+      if (!callback) callback = this[name];
+      var router = this;
+      Backbone.history.route(route, function(fragment) {
+
+        newrelic.setCustomAttribute('route', name);
+        var UUID = "uuid-"+(new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
```
kentmw: Should these overrides be mixins?

modules/Router.js
```diff
@@ -15,5 +15,48 @@
...
+        Logger.track({
+          UUID : UUID,
+          type : "routeChange",
+          state: "start",
+          route: name,
+          time: Date.now(),
+        });
+
+        var args = router._extractParameters(route, fragment);
```
kentmw: For each of these overrides, we need to make chainable, generic hooks for any logic (including logging). Then it would be worth overriding.

catbieber:
In terms of isolating the new relic code, the event tracking should be general enough that someone could replace the new relic integration with a google analytics integration, some other app, or their own custom service, without having to go digging in the logger itself. It should be a line or a couple lines when including/configuring the "logger".

We also probably need some updates to the documentation to explain what events are available and how to use this.

modules/Router.js
cmchin: I'm not sure what you mean by 'chainable, generic hooks'. Can you give me an example?

modules/Router.js
cmchin: Can you elaborate on this? I'm not sure how to make these overrides mixins. For a number of the overridden methods, I'm modifying the callback functions themselves, so I'm changing internal sections of the overridden functions. Is there a way to do this (with or without using mixins) that doesn't require copying and pasting methods from backbone?

modules/View.js
cmchin: In delegateEvents, I'm modifying the callback functions that are assigned to certain event triggers. delegateEvents is only called after every render to tie user actions to callback functions, but the function themselves are not called by delegate events, they're fired in response to clicks or key presses. Wrapping delegateEvents won't give us information about the timing of events that happen in response to clicks/key presses. To track these events, I wanted to wrap the callback functions with before/after signals sent to logger. Is there a way I can do that without copying delegateEvents from backbone?

kentmw: @cmchin, I spoke to @mandragorn about this and if we're going to re-write backbone, we at least need to be able to add hooks for application code. If I, as a programmer using torso, want to add "before" and "after" logic on event callbacks on a view, then it would be worth overriding delegate events.

cmchin:
@kentmw @catbieber @mandragorn 
Code review comments have been addressed in the most updated commit 
- new relic is no longer a dependency
- logger is removed
-  there is minimal reuse of backbone/torso code
- signals are handled by event triggers

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+      }
+      return trackingInfo;
+  	},
+
+    /**
+    * generate and returns uuid
+    * @return {String} unique identifier
+    */
+    getUUID : function(){
+      return (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
```
kentmw: unless something special is needed, _.uniqueId() is preferred.

modules/View.js
```diff
@@ -56,6 +56,72 @@
       if (!options.noActivate) {
         this.activate();
       }
+      // this.updateDelegateEvents();
```
kentmw: no commented code gets merged

modules/View.js
```diff
@@ -56,6 +56,72 @@
...
+    * @param eventName {String} description of event type 
+    * @return modified mehtod with before/after event triggers
+    * @method trackEvents
+    */
+    trackEvents : function(method, eventName){
+      var self = this;
+      var methodCopy = method;
+      method = _.bind(function(){
+        var uuid = (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+        this.trigger('clickEventTiming', {
```
kentmw: this event should be called, 'pre-dom-event' or something as generic. Then add a listener by default (perhaps in view constructor) to fire an event start.

modules/View.js
```diff
@@ -56,6 +56,72 @@
...
+          uuid: uuid,
+          type: 'clickEvent',
```
kentmw: the type here is incorrect. DOM event callbacks can be on any trigger: 'click', 'hover', 'mouseup'. They're all DOM events though

modules/View.js
```diff
@@ -56,6 +56,72 @@
...
+          state: 'start',
+          eventName: eventName,
+          time: Date.now(),
+        });
+
+        methodCopy.call(self);
+
+        this.trigger('routetiming', {
```
kentmw: same here: 'post-dom-event-callback' or something like that. Trigger with the same uuid (like you've done) as the pre-dom-event-callback.  Also add the listener in the constructor.

modules/View.js
```diff
@@ -56,6 +56,72 @@
...
+
+    /**
+    * updates Backbone's delegateEvents to bind triggers to all event callback functions 
+    * Binds DOM events with the view using events hash.
+    * Also adds feedback event bindings
+    * functions in the View's list of events
+    * @method updateDelegateEvents
+    * @override
+    */
+    delegateEvents: function(events){
```
kentmw: check for missing spaces between () and {
```
function(events) {
```

modules/View.js
```diff
@@ -56,6 +56,72 @@
...
+      for (var key in eventsClone) {
+        var method = eventsClone[key];
+        if (!_.isFunction(method)) method = this[eventsClone[key]];
+        if (!method) continue;      
+        this.trackEvents = _.bind(this.trackEvents, this);
+        var methodNew = this.trackEvents(method, '');
+        eventsClone[key] = methodNew;
+      }
+
+      Backbone.View.prototype.delegateEvents.call(this, eventsClone);
```
kentmw: Nice. You're capturing it before the delegateEvents is fired and updating the parameter.

modules/View.js
```diff
@@ -93,30 +159,34 @@
...
     templateRender: function(el, template, context, opts) {
       // Detach just this view's child views for a more effective hotswap.
       // The child views will be reattached by the render method.
+      var pageName = null;
+      if(opts){
+        pageName = opts.pageName;
+      }
+
+      var uuid = (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+      this.trigger('templateRenderTiming', {
```
kentmw: same here. Name should be something like: 'pre-render' and 'post-render' and the payload doesn't need state. That should be inferred by which event you are listening to.

modules/View.js
```diff
@@ -93,30 +159,34 @@
...
      */
     templateRender: function(el, template, context, opts) {
       // Detach just this view's child views for a more effective hotswap.
       // The child views will be reattached by the render method.
+      var pageName = null;
+      if(opts){
+        pageName = opts.pageName;
+      }
+
+      var uuid = (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
```
kentmw: _.uniqueId

modules/torso.js
```diff
@@ -41,7 +41,7 @@
                  pollingMixin,
                  validation,
                  Model, NestedCell, NestedModel, FormModel,
-                 ServiceCell,
+                 ServiceCell, EventTracker,
```
kentmw: You probably need to update bundle.js and commonJsImports.js

modules/View.js
```diff
@@ -56,6 +56,72 @@
...
+
+    /**
+    * called by updateDelegateEvents to wrap callback functions of event calls
+    * with event triggers
+    * @param method {function} callback method of event
+    * @param eventName {String} description of event type 
+    * @return modified mehtod with before/after event triggers
+    * @method trackEvents
+    */
+    trackEvents : function(method, eventName){
```
kentmw: Consider renaming this to be more general and also indicate that this is not part of the public API with double "__". Something like: '__wrapDelegatedEvents' or '__triggerDomEventHelpers' or something that's less "tracking" - that's just one application of these triggered events

modules/Router.js
```diff
@@ -15,5 +15,39 @@
...
+      var routeName = route;
+      if (!_.isRegExp(route)) route = this._routeToRegExp(route);
+      if (_.isFunction(name)) {
+        callback = name;
+        name = '';
+      }
+      if (!callback) callback = this[name];
+      var callbackCopy = function(){
+        var uuid = (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+        this.trigger('routeTiming',
```
kentmw: See all the comments about these type of event names and payloads.

Note, the convention of triggered events are "-" separated (not camelCased) amd that they describe what has or about to happen.

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+      }
+    },
+
+    /**
+    * saves information about event and wraps start and end signals in 'session' 
+    * @param eventInfo {JSON object} contains tracking information (including state: 'start' or 'end', uuid, sessionID)
+    * @return trackingInfo {JSON object} contains information about event to be passed back in 'end' signal
+    * @method track
+    */
+  	track: function(eventInfo) {
```
kentmw: no tab

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+    _sessionID: null,
+    _ajaxSession: [],
+
+
+    initialize: function(options){
+      this.options = options;
+      if (options){
+        if ('debug' in options){
+          this.debug = options.debug;
+          this.allSessions = {};
```
kentmw: I'm still hazy about what the definition of a "session" in this class is.

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+                  this.closeSession();
+                }
+            }
+          }
+      }
+      else{
+        _.extend(trackingInfo, this.startEvent(eventInfo));
+      }
+      return trackingInfo;
+  	},
```
kentmw: tab

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+      return (new Date()).getTime().toString(16)+Math.floor(1E7*Math.random()).toString(16);
+    },
+
+    /**
+    * tracks information about the 'start' of events
+    * @param eventInfo {JSON object} contains tracking information including UUID and time
+    * @method startEvent
+    */
+    startEvent: function(eventInfo) {
+      var uuid = eventInfo.UUID;
```
kentmw: can you default this? 
```
var uuid = eventInfo.UUID || _.uniqueId()
```

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+    },
+
+    /**
+    * tracks information about the 'start' of events
+    * @param eventInfo {JSON object} contains tracking information including UUID and time
+    * @method startEvent
+    */
+    startEvent: function(eventInfo) {
+      var uuid = eventInfo.UUID;      
+      eventInfo.UUID = this.getUUID;
```
kentmw: this is returning a function, not the invocation.

modules/EventTracker.js
```diff
@@ -0,0 +1,187 @@
...
+      eventInfo.sessionID = this._sessionID; //check if event info has sessionID property and throw error
+      eventInfo.startTime = Date.now();
```
kentmw: you probably want to let someone override the time?
```
eventInfo.startTime = eventInfo.startTime || Date.now()

File: modules/EventTracker.js [hunk 38]
```diff
@@ -0,0 +1,187 @@
...
+  } else if (typeof exports === 'object') {
+    module.exports = factory(require('underscore'), require('backbone'), require('./ServiceCell'));
+  } else {
+    root.Torso = root.Torso || {};
+    root.Torso.EventTracker = factory(root._, root.Backbone, root.Torso.ServiceCell);
+  }
+}(this, function(_, Backbone, ServiceCell) {
+  'use strict';
+
+  var EventTracker = ServiceCell.extend({
```
kentmw: You created all these wonderful triggered events inside Model, Router, and View, but I can't see you listening to any of them to generate events from this class? That begs the question, how are you planning to listen to each model, view, and router in order to generate the events? As long as you only have one EventTracker and it's created by default, it's possible to have in the constructors of the Model, View, and Router to add an "on" that registers the events to the EventTracker.

In Torso, we have not made a module rely on the existence of an instance of another module, which is what the previous suggestion requires. Do you have thoughts on how to add the dependency on an instance of an EventTracker but still have EventTracker be optional?

kentmw:
Great work @cmchin! I think you did a good job reducing the rewriting backbone. There's a few changes are left in order for across-the-board value from your commit.

Can you create test cases for this commit?

modules/EventTracker.js
cmchin: I accidentally left EventTracker in this commit (it should be deleted). I spoke with @mandragorn  and @catbieber and we've concluded that it would be best to trigger all the timing events to Torso.Events so that any customized event tracker can listen to Torso.Events to capture all render/router/fetch timings. Do you have thoughts about this?

Would you still prefer to trigger each event on 'this', and add an 'on' to the constructor and register the events to Events?

kentmw: Triggering all events through a single "Events" channel would create a Pub-Sub-like system. The hard part about that is that I'd want to, in many situations, subscribe to events that both match an event and the instance that triggered it. But, it allows for ease of access to all triggered events and a centralized place to pick up all events of a certain type.

If we trigger on Torso.Events, then the instance firing the event needs to be included in the event.

I'm thinking of the use-case of wanting to attach my own "pre-event-dom-callback" and "post-event-dom-callback"s in a view. This should be able to be performed inside the view without the help of Torso.Events. Perhaps routing all actual events through Torso.Events, and separately invoking (default noop) methods in the view before and after.

Is it possible to create something like: this.listenTo(Torso.Events, 'my-event', instanceWhoCalledMyEvent)? Perhaps: Torso.Events.notify(this, 'my-event', instanceWhoCalledMyEvent)?

kentmw: Trying to think if you triggered the event on the instance and on Torso.Events, would that cause issues involving multiple triggered callbacks? The events would be triggered in two different places. If not, it might give us the flexibility to do instance-level subscribing, and application-level subscribing.

mandragorn: Probably a bad idea, but we could trigger 2 events, one on Events with and one on the view.

mandragorn: @kentmw that would only happen if Events is also posting all events that are generated by the view.

kentmw: @mandragorn, I'm having a hard time tracking what parts of my comments you are referring to. But, I think that we are in agreement that we probably should trigger two events, one on the instance (model, view, or router) and one on the Events. Listening to Events will let us attach system-level logic while listening to the instance is more granular. For instance triggering, would it make sense to have automatic callbacks, or will I have to add this.on('my-pre-event', this.preEvent) and this.on('my-post-event', this.postEvent) on each view?

cmchin:
@kentmw @mandragorn @catbieber 

Made all the changes mentioned above:
- changed payload of events
- got rid of EventTracker
- Events are triggered on the object itself (this). Each view/router/module listens for timing events on itself and triggers an event on Torso.Events

modules/Router.js
```diff
@@ -15,5 +15,64 @@
    * @class  Router
    * @author kent.willis@vecna.com
    */
-  return Backbone.Router.extend({});
+
+  var Router = Backbone.Router.extend({
+
+    constructor: function(){
+      Backbone.Router.apply(this, arguments);
+      this.__applyTimingEventListeners(['pre-route', 'post-route']);
```
cmchin: @kentmw is there a better way to add all event listeners without enumerating each one as strings in the constructor?

modules/Model.js
```diff
@@ -1,23 +1,69 @@
...
+     * Overrides constructor to apply timing event triggers to Torso.Events
+     * @method constructor
+     * @override
+     */
+    constructor: function() {
+      Backbone.Model.apply(this,arguments);
+      this.__applyTimingEventListeners(['pre-fetch', 'post-fetch']);
+    },   
+
+    fetch: function(options) {
```
kentmw: add ```@override``` and ```@method``` annotations to this

modules/Model.js
```diff
@@ -1,23 +1,69 @@
...
+      var uuid = _.uniqueId();
```
kentmw: I just want to double check that my suggestion about using _.uniqueId() is applicable to your usecase. From the documentation:
```
Generate a unique integer id (unique within the entire client session). Useful for temporary DOM ids.
```
It restarts to 1 when you reload the page.

modules/Model.js
```diff
@@ -1,23 +1,69 @@
...
   } else if (typeof exports === 'object') {
-    module.exports = factory(require('underscore'), require('backbone'), require('./pollingMixin'));
+    module.exports = factory(require('underscore'), require('backbone'), require('./pollingMixin'), require('./Events'));
   } else {
     root.Torso = root.Torso || {};
-    root.Torso.Model = factory(root._, root.Backbone, root.Torso.Mixins.polling);
+    root.Torso.Model = factory(root._, root.Backbone, root.Torso.Mixins.polling, root.Torso.Events);
   }
-}(this, function(_, Backbone, pollingMixin) {
+}(this, function(_, Backbone, pollingMixin, Events) {
```
kentmw: We spoke at last night's Torso meeting about this patch. The direction you've taken this patch is great. One big thing that came up is that Torso.Events isn't quite the right place to put the timing events. Chat with @arikwex and @mandragorn about potential alternatives.

modules/Router.js
kentmw: @cmchin, I can't think of one. Although, I could see you passing in pairs of strings to which it can handle Events-specific logic across events (passing information from the start to the stop if needed).

I like being declarative about which events we are binding to timing events, at least for now.

modules/Router.js
```diff
@@ -15,5 +15,64 @@
...
+      if (!callback) callback = this[name];
+      var callbackCopy = function(){
+        var uuid = _.uniqueId();
+        this.trigger('pre-route', 
+          { uuid: uuid,
+            route: routeName, 
+            time: Date.now(),
+          });
+        callback.call(this);
+        this.trigger('post-route',
```
kentmw: the post-route should definitely have the routeName as well.

modules/Model.js
```diff
@@ -1,23 +1,69 @@
...
+    constructor: function() {
+      Backbone.Model.apply(this,arguments);
+      this.__applyTimingEventListeners(['pre-fetch', 'post-fetch']);
+    },   
+
+    fetch: function(options) {
+      var uuid = _.uniqueId();
+      this.trigger('pre-fetch', {
+        uuid: uuid,
+        time: Date.now(),
```
kentmw: I would add the options as part of the payload (both pre and post fetch)

modules/View.js
```diff
@@ -52,10 +52,43 @@
       this.feedbackModel = new Cell();
       this.__childViews = {};
       this.__feedbackEvents = [];
+      this.__applyTimingEventListeners(['pre-dom-event-callback', 'post-dom-event-callback', 'pre-render', 'post-render']);
```
kentmw: It seems strange we don't have to pair these up.

modules/View.js
```diff
@@ -93,30 +126,31 @@
...
     /**
      * Hotswap rendering system reroute method.
+     * triggers templateRenderTiming on view to track render times
      * @method templateRender
      * See Torso.templateRenderer#render for params
      */
     templateRender: function(el, template, context, opts) {
       // Detach just this view's child views for a more effective hotswap.
       // The child views will be reattached by the render method.
+      var pageName = null;
```
kentmw: Move the comment about detaching child view to above where you detach child views.

modules/View.js
```diff
@@ -93,30 +126,31 @@
...
+      if(opts){
+        pageName = opts.pageName;
```
kentmw: could you also check if pageName is in on this.pageName and use that.
```
opts = opts || {};
var pageName = opts.pageName || this.pageName;
```

modules/View.js
```diff
@@ -93,30 +126,31 @@
...
-    delegateEvents: function() {
-      Backbone.View.prototype.delegateEvents.call(this);
-      this.__generateFeedbackBindings();
-      this.__generateFeedbackModelCallbacks();
-      _.each(this.__childViews, function(view) {
-        if (view.isAttachedToParent()) {
-          view.delegateEvents();
-        }
+      this.trigger('post-template-render-timing', {
+        uuid: uuid,
```
kentmw: also add pageName

modules/View.js
```diff
@@ -441,6 +475,55 @@
 
     /************** Private methods **************/
 
+
+    /**
+    * sends event triggers to Torso.Events
+    * @param array of timing event names
+    * @method applyTimingEventListeners
+    */
+    __applyTimingEventListeners: function(timingEvents) {
```
kentmw: Not a big fan of having this repeated in every module. Can this be in whatever "timing events" module you create after speaking to @mandragorn and @arikwex?

modules/View.js
```diff
@@ -441,6 +475,55 @@
...
+    * @method __wrapDelegatedEvents
+    */
+    __wrapDelegatedEvents: function(method, eventName) {
+      var self = this;
+      var methodCopy = method;
+      method = _.bind(function(){
+        var uuid = _.uniqueId();
+        this.trigger('pre-dom-event-callback', {
+          uuid: uuid,
+          eventName: eventName,
```
kentmw: also add method

modules/View.js
```diff
@@ -441,6 +475,55 @@
...
+          time: Date.now(),
+        });
+
+        methodCopy.call(self);
+
+        this.trigger('post-dom-event-callback', {
+          uuid: uuid,
```
kentmw: include method and eventName in this as well.

torso-bundle.js
```diff
@@ -39,7 +39,66 @@
    * @class  Router
    * @author kent.willis@vecna.com
    */
-  return Backbone.Router.extend({});
+
+  var Router = Backbone.Router.extend({
+
+    constructor: function(){
```
kentmw: recreate bundle before pushing. Looks like it has old code in it.

kentmw:
@cmchin super close and really useful :+1:

modules/Model.js
cmchin: I'm not sure if  _.uniqueId() is appropriate in this case. Would it be a problem if several people were using the app at the same time? Would we be able to match the correct pre/post event triggers?

modules/View.js
cmchin: By 'these' do you mean the the pre and post events? What are your thoughts on how to pair them up?

modules/Model.js
kentmw: this is per-client, so for one user you won't have repeats unless they refreshed the browser

modules/View.js
kentmw: if these are only for timing events, maybe we could simplify it and say that you pass in 'dom-event-callback' and 'render' and we do the prefixing?

modules/Model.js
mandragorn: I don't think uniqueId will work because you want unique ids server-side, not client-side, correct?  Unless you uniquify them by adding some kind of unique session id.

kentmw: Yea, the thing is though, you can't really trust the uniqueness of something coming from the client side especially if its based on time (concurrent users, etc.). We can make it unique for each session on the browser and have the server generate true unique ids based on time, user information, and the browser session id. Or we could make psuedo-unique numbers on the client.

modules/View.js
cmchin: by this do you mean the name of the method itself in the code?

kentmw:
@mandragorn , @cmchin I'm excited about this pull request. Any status on the last touches?

kentmw:
@cmchin do you want to finish this patch? If I don't hear from you in a week, I'll just finish it for you.

PR #107, opened 2015-08-12
Code for issues 101, and 100

patrickckelly:
I added defaults for ListView's initialize so that if you extend the initialize method you don't need to attach things to this and put them in an arguments object. I added a check to the View's super method that allows you to create the ViewState before super is called and it doesn't get overridden.  I also added the activate method to the ListView's initialize.

### DIFF
--- a/modules/ListView.js
+++ b/modules/ListView.js
@@ -144,6 +144,7 @@
       this.super();
       this.listViewSetup(args);
       this.render();
+      this.activate();
     },
 
     /**
--- a/modules/ListView.js
+++ b/modules/ListView.js
@@ -108,7 +108,7 @@
    */
   var ListView = View.extend({
     className: '',
-    _collection: null,
+    collection: null,
     _modelName: '',
     _childView: null,
     _modelToViewMap: null,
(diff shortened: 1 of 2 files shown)


mandragorn:
+1 - I want to make sure @kentmw reviews it before merging.

modules/ListView.js
```diff
@@ -156,10 +157,10 @@
...
       args = args || {};
       this._modelName = args.childModel || 'model';
-      this._collection = args.collection;
-      this._childView = args.childView;
-      this._template = args.template;
-      this._childrenContainer = args.childrenContainer;
+      this.collection = args.collection || this.collection;
+      this._childView = args.childView || this.childView;
+      this._template = args.template || this.template;
+      this._childrenContainer = args.childrenContainer || this.childrenContainer;
```
mandragorn: These defaults should have the same visibility in both places.  For backwards compatibility I'd suggest:
  this._childView = args.childView || this._childView

These are different because there isn't already a backbone concept for them (like collection).

mandragorn: that is the same for childView, template and childrenContainer

PR #108, opened 2015-08-13
Upgrading to jsdom 6.1.0

mandragorn:
Just testing to see what happens with continuous integration.  I was fairly quickly able to setup io.js and am able to switch between io.js and node.js locally in a clean way using n (https://github.com/tj/n)

### DIFF
--- a/modules/collectionRegistrationMixin.js
+++ b/modules/collectionRegistrationMixin.js
@@ -232,8 +232,10 @@
         // Create a new request list
         for (requesterIdx = 0; requesterIdx < requesterLength; requesterIdx++) {
           storedIds = this.getRequesterIds(requesters[requesterIdx]);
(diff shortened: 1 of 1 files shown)


kentmw:
@mandragorn I hear that io.js and node.js are merging again? http://thenextweb.com/dd/2015/06/16/node-js-and-io-js-are-settling-their-differences-merging-back-together/

mandragorn:
correct, they just haven't now.  Either way I'm not sure it buys us anything to use io.js right now since it still didn't work out-of-the-box (unless I configured something wrong).  For now I'm going to let this sit while Mike looks into it.

mikeobr:
io.js and node merged yesterday: https://nodejs.org/en/blog/release/v4.0.0/, I'm going to play around with upgrading.

PR #136, opened 2015-10-29
Adding injectGlobalView to address #135.

### LINKED ISSUE #135: How does a parent view inject other views that are not instantiated by the parent
As a developer I would like to inject a global view (one that isn't instantiated by the parent) when the parent renders.  But this view should not be disposed when the parent is disposed since its lifecycle is broader than the parent.

For example the parent view is a perspective and the global view is the footer.  The footer view I want to use on a bunch of different perspectives so I only want to instantiate it once.  That said, I don't want the view to be disposed if the parent view is disposed (since I will want to attach it to the next perspective).

mandragorn:
fixed #135


modules/View.js
```diff
@@ -264,15 +270,122 @@
...
+    /**
+     * Registers the child view if not already done so, then calls view.attach with the element argument
+     * @param $el {jQuery element} the element to attach to.
+     * @param view {View} the child view
+     * @param global {Boolean} The view is a global view instead of a child view
+     *                         (global views are not disposed when the parent is disposed)
      * @method attachChildView
      */
-    attachChildView: function($el, view) {
+    attachChildView: function($el, view, global) {
```
mandragorn: remove reference to global

kentmw:
Not a big fan of the duplication of methods/logic between child/shared views but i love the distinction and the functionality this brings.

Please add this to master as well!!

modules/ListView.js
```diff
@@ -59,10 +59,10 @@
       var childView = this.getChildView(model);
       if (childView) {
         childView.dispose();
-        this.unregisterChildView(childView);
+        this.unregisterTrackedView(childView, { shared: false });
```
kentmw: all these places you have shared:false are just bookkeeping right? it's not necessary?

mandragorn: correct, it is just being explicit, but not required.  unregisterTrackedView(childView); will treat the view as a child, not shared, view.

kentmw:
+1 please make a pull request for master

torso-bundle.js
```diff
@@ -3523,10 +3727,10 @@
         injectionSite = this.$el;
       args = args || {};
       this._modelName = args.childModel || 'model';
-      this._collection = args.collection;
-      this._childView = args.childView;
-      this._template = args.template;
-      this._childrenContainer = args.childrenContainer;
+      this.collection = args.collection || this.collection;
```
mandragorn: I wasn't intending on including this, but the jasmine tests pass, any issues with including this in this pull request?

mandragorn: This lines the .collection property of backbone up with the collection associated with the ListView.

Issue #110, opened 2015-08-22
collection.url in registrationCollectionMixin should use _.result


kentmw:
```
          return $.ajax({
              type: collection.fetchHttpAction,
              url: collection.url + collection.getByIdsUrl,
              contentType: 'application/json; charset=utf-8',
              data: JSON.stringify(idsToFetch)
            })
```
uses collection.url straight and doesn't allow for a method

kentmw:
+1 had this happen to me again... My function returning a url failed.

mandragorn:
Fix in both 0.1.x and master branches.

kentmw:
https://github.com/vecnatechnologies/backbone-torso/pull/138

PR #138, opened 2015-11-24
#110 functions for collection urls are now supported.

blakejwc:


kentmw:
+1 Thanks @blakejwc for adding this!

PR #147, opened 2015-12-31
List view defaults

mandragorn:


torso-bundle.js
```diff
@@ -488,7 +488,7 @@
           }
           return $.ajax({
               type: collection.fetchHttpAction,
-              url: collection.url + collection.getByIdsUrl,
+              url: _.result(collection, 'url') + collection.getByIdsUrl,
```
kentmw: while i want this change very much, you have it just in the bundle. Mind adding it to the collection mixin file as well?

mandragorn: It is in the collection mixin.  The bundle was created just before pushing.
The person who made the change to the mixin didn't bundle before comitting

Issue #144, opened 2015-12-30
Update to newest backbone version


kentmw:
We're about a year old as of me writing this. Time to upgrade.

kentmw:
http://backbonejs.org/#changelog

Issue #141, opened 2015-12-10
ListView does not follow the Torso life cycle


kentmw:
Particularly it doesn't inject the child views and so they are not marked as attached to the parent nor are any of the attach callbacks invoked. Also because it doesn't call injectView, the child views are unrendered at the beginning.

Our purposed solution is to move the last part of the listview's render and make it mirror that of calling injectView for each child view that it produced, with the exception of dom generation which it will do all in one step. See https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/ListView.js#L214-L222 for the section that needs to be updated. These lines should be refactored another method and replicate the attachView for each child view. The notable difference is that the dom fragments are created and inserted into the injection site. Then we need to loop over the child views, detatching them (which means removing all the dom as one, and going through each child view and processing the detach logic. See: https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/View.js#L177-L179), reregister the child views, then call render on each one, then creating the dom fragment with all of the child views dom, and for each child view: invoke the attachCallbacks and then set their attachedToParent to true, and then activate it (if the list view was passed activateChildView option during initialize).

kentmw:
I haven't checked but I believe this is another symptom. It should be tested after this issue is complete (or at least kept in mind while solving this issue)

I have two nested list views: One list view has a child view that contains a list view. When I render on the parent list view, the events on the inner list view child views are not delegated. I have to add:
```
  render: function() {
    Torso.ListView.prototype.render.call(this);
    this.renderChildViews();
  }
```
to both the outer and inner list views to get it to delegate properly

kentmw:
Taking a hit in performance, this now works

Issue #121, opened 2015-09-03
ListView's child views are never rendered


kentmw:
Because we removed the rendering on initialization and moved it to when you attach the view to the DOM, ListView's child views never get rendered. This is because the child views are never "attached" - they are appended to each other in a fragment and placed onto the page at once.

PR #148, opened 2015-12-31
#141, #144, #121 - list view's child views now follow torso lifecycles

kentmw:
Also upgraded backbone version to 2.1.3


kentmw:
NOTE: there is a performance hit with this. We are rendering child views on list view renders

kentmw:
Perhaps we should make it so sorting events don't cause a re-renders on the child views before we merge this.

modules/ListView.js
```diff
@@ -319,15 +324,23 @@
...
      * @return a DOM fragment with child view elements appended
-     * @method __buildChildViewsFragment
+     * @method __emptyAndRebuildChildViewsFragment
      * @private
      */
-    __buildChildViewsFragment: function(renderAlso) {
+    __emptyAndRebuildChildViewsFragment: function(renderAlso) {
       var injectionFragment = document.createDocumentFragment();
+      // Clearing the DOM will reduce the repaints needed as we detach each child view.
+      this.$el.empty();
```
mandragorn: this feels like it should live outside of this method.  It is mutating the parent, where this method is about generating the child content.

modules/ListView.js
```diff
@@ -319,15 +324,23 @@
...
      _.each(this.modelsToRender(), function(model) {
         var childView = this.getChildViewFromModel(model);
         if (childView) {
+          childView.detach();
```
mandragorn: oh, right, this mutates the parent.  nvm.

mandragorn:
+2

PR #158, opened 2016-01-01
Moved all torso modules to use constructors instead of initialize

kentmw:


modules/ListView.js
```diff
@@ -171,7 +170,8 @@
      *   @param [args.modelId='cid'] {String} - model property used as identifier for a given model. This property is saved and used to find the corresponding view.
      *   @param [args.childModel='model'] {String} - name of the model argument passed to the child view during initialization
      */
-    initialize: function(args) {
+    constructor: function(args) {
```
mandragorn: note that this requires an upgrade.  If a view set its collection before calling the ListView initialize it will now either have to convert to constructors or use setCollection(); - just make sure this gets into the release notes.

mandragorn:
+2

modules/ListView.js
kentmw: The three ways you set the collection on a list view:
1. through argument on intialization:  the constructor handles this well.
2. by setting it after initialization: use setCollection()
3. by defining it as part of the extension of the ListView: in this case this.collection should be properly referenced by the constructor

Am I missing something? I'm trying to figure out when things would go wrong. How besides defining it as part of the extension of the list view would a view set the collection before calling initialize?

Issue #159, opened 2016-01-02
ListView's addChildView doesn't respect child view's lifecycle


kentmw:
It is still missing activation and attach logic

PR #160, opened 2016-01-02
#159 - listview's addChildView now activates/attaches new child view

kentmw:


modules/View.js
```diff
@@ -198,13 +198,24 @@
 
     /**
      * If detached, will replace the element passed in with this view's element and activate the view.
-     * @param $el [jQuery element] the element to attach to. This element will be replaced will this view
+     * @param $el {jQuery element} the element to attach to. This element will be replaced will this view
```
mandragorn: will this view -> with this view

modules/View.js
```diff
@@ -198,13 +198,24 @@
...
+     * @param [options] {Object} optional options
+     * @param   [options.replaceMethod] {Fucntion} if given, this view will invoke replaceMethod function
+     *                                             in order to attach the view's DOM to the parent instead of calling $el.replaceWith
+     * @param   [options.discardInjectionSite=false] {Booleon} if set to true, the injection site is not saved.
      * @method attach
      */
-    attach: function($el) {
+    attach: function($el, options) {
```
mandragorn: unfortunate $el arg needs to be kept for backwards compatability.  Could we look at the object to determine whether it is providing a replaceMethod or not?  Then have a single polymorphic argument?

modules/View.js
```diff
@@ -353,7 +364,7 @@
       options = options || {};
       view.detach();
       this.registerTrackedView(view, options);
-      view.attach($el);
+      view.attach($el, options);
```
mandragorn: oh, but then this doesn't work as well (in response to my previous comment).

modules/View.js
kentmw: they're both objects, so we'd have to duck-type identify the element

modules/View.js
kentmw: If we wanted to duck-type identify it in attach, then we could construct the options object here that contains the $el.
```
  options.$el = $el;
  view.attach(options)
```

mandragorn:
+2 - might want to think about the options thing, but I think with the attachView use case this is simplest.

Could do this, but I'm not sure its worth the complexity:

3 states of the replace method logic:
1) params: $el - do the original logic
2) params: $el, options (with replace method) - do the new logic
3) params: options (with replace method and $el) - do original logic but pass in options.$el as first param.
4) params: options (with $el and no replace method) - do same as 1)

modules/View.js
```diff
@@ -198,13 +198,24 @@
 
     /**
      * If detached, will replace the element passed in with this view's element and activate the view.
-     * @param $el [jQuery element] the element to attach to. This element will be replaced will this view
+     * @param $el {jQuery element} the element to attach to. This element will be replaced with this view
```
mandragorn: this is a conditionally optional parameter - update docs.

kentmw:
Kent Willis (kentmw@gmail.com)
to me, it comes down to if we can appropriately and safely duck type $el
otherwise we don't know if it's $el or options passed in
Josh Young (jyoung@vecna.com)
right, I think we can do it just by testing for replaceMethod.
unless $el has a replaceMethod field
Kent Willis (kentmw@gmail.com)
but disgardInjectionSite could be used without replaceMethod
also, that doesn't leave us much for extending the  options set
Josh Young (jyoung@vecna.com)
true
yea, I'm not sure its worth the complexity.
null arguments bother me
but I think its fine here.
Kent Willis (kentmw@gmail.com)
i'm was thinking of checking for $el.jquery
but that's not very extensible for other frameworks
Josh Young (jyoung@vecna.com)
yea, that doesn't seem great.
Kent Willis (kentmw@gmail.com)
and then other ways just don't seem right (attr, html, etc.)
okay, we leave it?
Josh Young (jyoung@vecna.com)
yea, leave it. Update the attach doc to method that $el is optional

mandragorn:
+2

Issue #165, opened 2016-01-03
ListView's sort method is much slower now


kentmw:
After 0.3.0 when the list view respects child view's lifecycle, the sort and reset methods are slow

PR #168, opened 2016-01-03
#165 - sort now doesn't rerender but moves elements - much faster. Al…

kentmw:
…so fixed misleading teset cases for speed concerning list view update method (collection.reset). Also fixed dependency issues with adding NestedCell as viewState


modules/ListView.js
```diff
@@ -430,7 +464,8 @@
      */
     __generateChildArgs: function(model) {
       var args = {
-        'context': _.extend({}, _.result(this, '__childContext'))
+        'context': _.extend({}, _.result(this, '__childContext')),
+        'listView': this
       };
```
mandragorn: where is this used?

kentmw: no where by torso. But if you want a child view to listen to events like "reorder" and other listview communication, i thought it couldn't hurt and could be very useful

modules/ListView.js
```diff
@@ -430,7 +464,8 @@
      */
     __generateChildArgs: function(model) {
       var args = {
-        'context': _.extend({}, _.result(this, '__childContext'))
+        'context': _.extend({}, _.result(this, '__childContext')),
+        'listView': this
```
mandragorn: Add this to the documentation somewhere.

mandragorn:
+1 - just needs clear documentation on the parameters provided to a child view when it is created.

mandragorn:
+2

Issue #166, opened 2016-01-03
ListView's sort method is much slower now


kentmw:
Now after 0.3.0 when ListView respects child view's lifecycles

PR #173, opened 2016-01-05
#166 - list view reset with a few models is much faster

kentmw:


modules/ListView.js
```diff
@@ -349,10 +366,32 @@
...
+      var view = this,
+        renderNeeded = false,
+        oldViews = this.getItemViews(),
+        newViews = this.__createItemViews(),
+        staleViews = this.__getStaleItemViews(),
+        sizeOfOldViews = _.size(oldViews),
+        sizeOfNewViews = _.size(newViews),
+        sizeOfStaleViews = _.size(staleViews),
+        changes = sizeOfNewViews + sizeOfStaleViews,
+        percentChange = changes / Math.max((sizeOfOldViews - sizeOfStaleViews + sizeOfNewViews), 1),
```
mandragorn: Readability:

    sizeOfFinalViews = sizeOfOldViews - sizeOfStaleViews + sizeOfNewViews
    percentChange = changes / Math.max(sizeOfFinalViews, 1);

modules/ListView.js
```diff
@@ -349,10 +366,32 @@
...
+        threshold = this.updateThreshold || 0.5;
+      if (!changes) {
+        this.reorder();
+        return;
+      }
+      // A switch from empty to not empty or vise versa, needs a new render
+      renderNeeded = (!sizeOfOldViews && sizeOfNewViews) || (sizeOfOldViews && sizeOfOldViews === staleViews && !sizeOfNewViews);
```
mandragorn: break up this logic into separate lines for easier readability.

modules/ListView.js
```diff
@@ -349,10 +366,32 @@
...
+      if (renderNeeded || percentChange >= threshold) {
```
mandragorn: make the percentChange part of the lines leading up to rederNeeded (since its part of that decision).  Then just if (renderNeeded) here.

modules/ListView.js
```diff
@@ -349,10 +367,31 @@
...
+        sizeOfStaleViews = _.size(staleViews),
+        changes = sizeOfNewViews + sizeOfStaleViews,
+        percentChange = changes / Math.max((sizeOfOldViews - sizeOfStaleViews + sizeOfNewViews), 1),
+        threshold = this.updateThreshold || 0.5;
+      if (!changes) {
+        this.reorder();
+        return;
+      }
+      // A switch from empty to not empty or vise versa, needs a new render
+      renderNeeded = (!sizeOfOldViews && sizeOfNewViews) || (sizeOfOldViews && sizeOfOldViews === sizeOfStaleViews && !sizeOfNewViews);
```
mandragorn: Readability: separate this into separate lines.

modules/ListView.js
```diff
@@ -349,10 +367,31 @@
...
+      if (renderNeeded || percentChange >= threshold) {
```
mandragorn: Readability: make this part of the separate lines leading up to renderNeeded.  This line should just be if (renderNeeded).

My rule of thumb is one action per line.

modules/ListView.js
```diff
@@ -349,10 +367,31 @@
...
+        this.$el.empty(); // TODO find out if this is usefull...
+        _.each(staleViews, function(staleViewInfo) {
+          _removeItemView.call(view, staleViewInfo.view, staleViewInfo.modelId);
+        });
         this.__delayedRender();
+      } else {
+        this.__updateByAddingRemoving(oldViews, newViews, staleViews);
```
mandragorn: this is the remaining case where the %change < threshold?

modules/ListView.js
```diff
@@ -366,31 +405,74 @@
...
+      return !_.isEmpty(this.getItemViews());
+    },
+
+    /**
+     * @returns {Array of views} Returns unordered list of views generated by this list view
+     * @method getItemViews
+     */
+    getItemViews: function() {
+      var view = this;
+      var orderedViewIds = _.map(this.__orderedModelIdList, function(modelId) { return view.__modelToViewMap[modelId]; });
```
mandragorn: too much going on in this line :).

is there a method already that takes a modelId and returns the view id?  Just pass that into _.map() and I think map also has a convenience for the context.

    _.map(this.__orderedModelIdList, this.getViewIdFromModelId, this);

or something like that.

modules/ListView.js
```diff
@@ -366,31 +405,74 @@
...
+      return _.map(orderedViewIds, this.getChildView, this);
```
mandragorn: yea! like this...

modules/ListView.js
```diff
@@ -366,31 +405,74 @@
...
+    },
+
     /************** Private methods **************/
 
     /**
-     * Creates a new child view if there doesn't exist one for a model
-     * @method __createChildViews
+     * Creates a new item view for a model if there doesn't exist one
```
mandragorn: update doc to the plural

modules/ListView.js
```diff
@@ -366,31 +405,74 @@
...
-      var removedChildViews = [];
+    __createItemView: function(model, noUpdateToIdList) {
+      var itemView,
+        ItemViewClass = this.childView;
+      if (!_.isFunction(this.childView.extend)) {
+        ItemViewClass = this.childView(model);
+      }
+      itemView = new ItemViewClass(this.__generateChildArgs(model));
+      this.registerTrackedView(itemView, { shared: false });
+      this.__modelToViewMap[model[this.__modelId]] = itemView.cid;
```
mandragorn: shouldn't this be model.get(this.__modelId) ?  Instead of model[this.__modelId].

Oh I keep forgetting this is the name of the model id, like cid or something else?  What else would you reference again that is useful?  id?  Is there any other option?

modules/ListView.js
```diff
@@ -366,31 +405,74 @@
...
+    /**
+     * Gets all child views that have models that are no longer tracked by modelsToRender
+     * @method __getStaleItemViews
+     * @private
+     */
+    __getStaleItemViews: function() {
+      var staleChildViews = [];
       var modelsWithViews = _.clone(this.__modelToViewMap);
       _.each(this.modelsToRender(), function(model) {
         var childView = this.getChildViewFromModel(model);
```
mandragorn: change to itemView here and below?

modules/ListView.js
```diff
@@ -401,13 +483,12 @@
...
         if (childView) {
-          removedChildViews.push(childView);
-          _removeChildView.call(this, childView, modelId);
+          staleChildViews.push({ view: childView, modelId: modelId });
         } else {
           delete this.__modelToViewMap[modelId];
         }
       }, this);
-      return removedChildViews;
+      return staleChildViews;
```
mandragorn: staleItemViews

modules/ListView.js
```diff
@@ -431,27 +512,71 @@
...
+    },
+
+    /**
+     * Updates the internal list of model ids that correspond to the models used for the current
+     * list of item views. The order is the same order of the item views.
+     * @method __updateOrderedModelIdList
+     * @private
+     */
+    __updateOrderedModelIdList: function() {
+      this.__orderedModelIdList = _.pluck(this.modelsToRender(), this.__modelId);
```
mandragorn: this is really a snapshot of the model order at this time?  Maybe that would be a clearer naming?  orderedModelIdList looks like it should be a live reflection of modelsToRender()'s order.  Which it isn't (it just only valid once this method is called if modelsToRender() changes).

test/spec/functional/listView.js
```diff
@@ -398,6 +404,84 @@ describe('A List View', function() {
...
 
+  it('can reset a few of many models correctly and in a reasonable time', function() {
+    var startTime, endTime, i,
+      models = [],
+      numberOfViews = 1000,
+      threshold = 500;
+    for (i = 0; i < numberOfViews; i++) {
+      models.push(new Model())
+    }
+    myListView.dispose();
```
mandragorn: why define list view globally if you always dispose it.  Seems like extra work, instead of a list view inside of each it method.

test/spec/functional/listView.js
```diff
@@ -398,6 +404,84 @@ describe('A List View', function() {
...
+    myListView.attach($('body'));
+    myCollection.reset(models);
+    startTime = new Date().getTime();
+    myCollection.reset(models.slice(0, numberOfViews - 100));
+    endTime = new Date().getTime() - startTime;
+    console.log('Removed 100 old views at the end of ' + numberOfViews + ' views using reset in ' + endTime + 'ms');
+    expect(myListView.getItemViews().length).toBe(numberOfViews - 100);
+    expect(endTime < threshold).toBe(true);
+
+    startTime = new Date().getTime();
```
mandragorn: separate test case.

test/spec/functional/listView.js
```diff
@@ -398,6 +404,84 @@ describe('A List View', function() {
...
+    myCollection.reset(models.slice(100, numberOfViews - 100));
+    endTime = new Date().getTime() - startTime;
+    console.log('Removed 100 old views at the front of ' + (numberOfViews - 100) + ' views using reset in ' + endTime + 'ms');
+    expect(myListView.getItemViews().length).toBe(numberOfViews - 200);
+    expect(endTime < threshold).toBe(true);
+
+    var newModels = []
```
mandragorn: separate test case.

test/spec/functional/listView.js
```diff
@@ -398,6 +404,84 @@ describe('A List View', function() {
...
+    endTime = new Date().getTime() - startTime;
+    console.log('Added 100 new views at the front of ' + (numberOfViews - 200) + ' views using reset in ' + endTime + 'ms');
+    expect(myListView.getItemViews().length).toBe(numberOfViews - 100);
+    expect(myListView.getItemViews()[0].model.cid).toBe(myCollection.at(0).cid);
+    expect(myListView.getItemViews()[numberOfViews - 101].model.cid).toBe(myCollection.at(numberOfViews - 101).cid);
+    expect(endTime < threshold).toBe(true);
+
+    threshold = 1000;
+
+    myListView.dispose();
```
mandragorn: new test case.

test/spec/functional/listView.js
```diff
@@ -398,6 +404,84 @@ describe('A List View', function() {
...
+    startTime = new Date().getTime();
+    myCollection.reset(newModels);
+    endTime = new Date().getTime() - startTime;
+    console.log('A list view using a children container: added 100 to the front and removed 200 views to ' + numberOfViews + ' item views using reset in ' + endTime + 'ms');
+    expect(myListView.getItemViews().length).toBe(numberOfViews - 100);
+    expect(myListView.getItemViews()[0].model.cid).toBe(myCollection.at(0).cid);
+    expect(myListView.getItemViews()[numberOfViews - 101].model.cid).toBe(myCollection.at(numberOfViews - 101).cid);
+    expect(endTime < threshold).toBe(true);
+
+    _.shuffle(models);
```
mandragorn: new test case.

the test cases can share a describe and before if you want to share setup.

modules/ListView.js
kentmw: yea, i updated the doc. it's usually always id or cid

kentmw:
https://github.com/vecnatechnologies/backbone-torso/pull/174/files

PR #174, opened 2016-01-05
#166 - list view reset with a few models is much faster

kentmw:


modules/ListView.js
```diff
@@ -233,10 +250,10 @@
 
       this.collection = collection;
 
-      this.listenTo(this.collection, 'remove', removeChildView);
-      this.listenTo(this.collection, 'add', addChildView);
-      this.listenTo(this.collection, 'sort', this.reorder);
-      this.listenTo(this.collection, 'reset', this.update);
+      this.listenTo(this.collection, 'remove', removeItemView, this);
```
mandragorn: I don't think you need the last this param.  That should already bind to this by default (since it is the same this that listenTo is called from).

modules/ListView.js
```diff
@@ -293,8 +310,19 @@
...
      */
     reorder: function() {
-      var firstChildView, elements = [];
-     _.each(this.modelsToRender(), function(model, index) {
+      var firstChildView, sameOrder,
+        elements = [],
+        models = this.modelsToRender(),
+        newOrderOfIds = _.pluck(models, this.__modelId);
+      if (_.size(newOrderOfIds) === _.size(this.__orderedModelIdList)) {
+        sameOrder = _.reduce(this.__orderedModelIdList, function(result, oldId, index) {
```
mandragorn: there is a lot going on in these lines.  Took a couple of reads to understand.

modules/ListView.js
```diff
@@ -293,8 +310,19 @@
      * @method reorder
      */
     reorder: function() {
-      var firstChildView, elements = [];
-     _.each(this.modelsToRender(), function(model, index) {
+      var firstChildView, sameOrder,
+        elements = [],
+        models = this.modelsToRender(),
+        newOrderOfIds = _.pluck(models, this.__modelId);
+      if (_.size(newOrderOfIds) === _.size(this.__orderedModelIdList)) {
```
mandragorn: explicitly call out sameSize as a var.  Use it here and on line 322 instead of hiding it with sameOrder.

modules/ListView.js
```diff
@@ -350,10 +379,31 @@
...
+        sizeOfNewViews = _.size(newViews),
+        sizeOfStaleViews = _.size(staleViews),
+        sizeOfFinalViews = sizeOfOldViews - sizeOfStaleViews + sizeOfNewViews,
+        changes = sizeOfNewViews + sizeOfStaleViews,
+        percentChange = changes / Math.max(sizeOfFinalViews, 1),
+        fromEmptyToNotEmpty = !sizeOfOldViews && sizeOfNewViews,
+        fromNotEmptyToEmpty = sizeOfOldViews && sizeOfOldViews === sizeOfStaleViews && !sizeOfNewViews,
+        threshold = this.updateThreshold || 0.5,
+        signficantChanges = percentChange >= threshold;
+      if (!changes) {
```
mandragorn: changes is an int, not boolean - changes <= 0

modules/ListView.js
```diff
@@ -367,48 +417,110 @@
...
-    __createChildViews: function() {
-      var newChildViews = [];
-      _.each(this.modelsToRender(), function(model) {
+    __createItemViews: function() {
+      var newItemViews = [];
+      _.each(this.modelsToRender(), function(model, indexOfModel) {
         var childView = this.getChildViewFromModel(model);
         if (!childView) {
-          newChildViews.push(this.__createChildView(model));
+          newItemViews.push({ view: this.__createItemView(model, true), indexOfModel: indexOfModel });
```
mandragorn: I find multiple properties easier to read if they are on separate lines as a block.  Easier to tell when one ends and the other starts, esp when there are method calls with parameters (extra commas).

modules/ListView.js
```diff
@@ -367,48 +417,110 @@
...
       _.each(modelsWithViews, function(viewId, modelId) {
-        var childView = this.getChildView(viewId);
-        if (childView) {
-          removedChildViews.push(childView);
-          _removeChildView.call(this, childView, modelId);
+        var itemView = this.getChildView(viewId);
+        if (itemView) {
+          staleItemViews.push({ view: itemView, modelId: modelId });
         } else {
           delete this.__modelToViewMap[modelId];
```
mandragorn: This is an unexpected side effect of this method.

modules/ListView.js
```diff
@@ -367,48 +417,110 @@
...
+     *     view: stale item view,
+     *     modelId: id of model item
+     *   }]} If provided, stale items will not be found, but this array will be used instead.
+     * @private
+     */
+    __removeStaleItemViews: function(staleItemViewInfo) {
+      var view = this;
+      staleItemViewInfo = staleItemViewInfo || this.__getStaleItemViews();
+      _.each(staleItemViewInfo, function(staleViewInfo) {
+        _removeItemView.call(view, staleViewInfo.view, staleViewInfo.modelId);
```
mandragorn: I would expect this to actually remove it from the _modelToViewMap

modules/ListView.js
```diff
@@ -432,27 +545,70 @@
...
+    __updateByAddingRemoving: function(oldViews, newViews, staleViews) {
+      var firstChildViewLeft, injectionSite,
+        view = this,
+        sizeOfOldViews = _.size(oldViews),
+        sizeOfNewViews = _.size(newViews),
+        sizeOfStaleViews = _.size(staleViews);
+      if (view.childrenContainer && sizeOfOldViews && sizeOfOldViews == sizeOfStaleViews) {
+        // we removed all the views!
+        injectionSite = $('<span>');
+        _.first(oldViews).$el.before(injectionSite);
```
mandragorn: so this adds the injection site before the old views?

Oh, so you can remove the old views?  You leave them attached so that they take advantage of the jquery move?

mandragorn:
-1 tests do not pass.

mandragorn:
see https://travis-ci.org/vecnatechnologies/backbone-torso/jobs/100399766 for test output.

modules/ListView.js
kentmw: should i remove it or update the docs? I just thought I'd clean up if somehow the state gets off

mandragorn: remove it.  If you want to clean up then do that explicitly.  Hidden state changes will only cause pain in the future.

modules/ListView.js
kentmw: Ah, i see the confusion. Yes, it does remove the entry in _modelToViewMap. It does so in _removeItemView.

The delete that happens in __getStaleItemViews removes ids in the map that don't have an assoicated view. This should never happen as that map should be managed carefully. Rather than throw an exception, i deleted the entry there

modules/ListView.js
mandragorn: if you rename the method and update the docs then that works too.  I don't really mind which.

modules/ListView.js
kentmw: the problem was that when you remove all the views that were put into the itemContainer, the injectionSite was not replaced back into the dom, so I have no idea where to stick the new views. So, before I yank out the last of the views, I stick in my own injectionsite beforehand.

test/spec/functional/listView.js
```diff
@@ -344,6 +347,7 @@ describe('A List View', function() {
       childModel: 'item',
```
mandragorn: should these change to modelName from childModel in the tests?

test/spec/functional/listView.js
```diff
@@ -344,6 +347,7 @@ describe('A List View', function() {
...
       childView: ItemView
```
mandragorn: Should childView -> itemView in tests?

test/spec/functional/listView.js
```diff
@@ -376,56 +381,121 @@ describe('A List View', function() {
...
+
+      beforeEach(function() {
+        myListView.dispose();
+        myListView = new MyListView({
+          collection: myCollection,
+          childModel: 'item',
+          childView: ItemView,
+          emptyTemplate: Handlebars.compile('<div class="empty-list"></div>'),
+          template: Handlebars.compile('<div class="templated-list"></div><div inject="children"></div>'),
+          childrenContainer: 'children'
```
mandragorn: itemContainer?

mandragorn:
:+1:   I've probably missed something, but I've run out of things to check complain about.

Issue #176, opened 2016-01-12
FormView has problems re-rendering select options because this.stickit()


mandragorn:
For some reason it matches up whitespace text nodes as the current node with option nodes as the new node during hotswap causing it to generate mis-matched option list:

First render is ok:

        <select class="visit-type__select-service-type" name="serviceType" data-model="serviceType" data-feedback="serviceType" id="service-type" aria-label="Select service type">
          
            
            
            
            
            
        <option value="defaultOption">Select Service</option><option value="CT Scan">CT Scan</option><option value="Diagnostic Mammogram">Diagnostic Mammogram</option><option value="Infectious Disease">Infectious Disease</option><option value="MRI">MRI</option><option value="Screening Mammogram">Screening Mammogram</option></select>
        <select class="visit-type__select-visit-type visit-type__select-service-type--disabled" name="visitType" data-model="visitType" data-feedback="visitType" id="visit-type" aria-label="Select Visit Type" disabled="">
          
        <option value="defaultOption">Select Visit Type (optional)</option></select>

Template generated for 2nd render is ok:

        <select class="visit-type__select-service-type"
                name="serviceType" data-model="serviceType" data-feedback="serviceType" name="service-type" id="service-type" 
                aria-label="Select service type">
          <option value="defaultOption">
            Select Service
          </option>
            <option value="CT Scan">
              CT Scan
            </option>
            <option value="Diagnostic Mammogram">
              Diagnostic Mammogram
            </option>
            <option value="Infectious Disease">
              Infectious Disease
            </option>
            <option value="MRI">
              MRI
            </option>
            <option value="Screening Mammogram">
              Screening Mammogram
            </option>
        </select>

But after template renderer merges, they are not ok:

        <select class="visit-type__select-service-type" name="serviceType" data-model="serviceType" data-feedback="serviceType" id="service-type" aria-label="Select service type">
          
            
            
            
            
            
        <option value="defaultOption">Select Service</option><option value="CT Scan">CT Scan</option><option value="Diagnostic Mammogram">Diagnostic Mammogram</option><option value="defaultOption">Infectious Disease</option><option value="Diagnostic Mammogram">MRI</option><option value="MRI">Screening Mammogram</option></select>
        <select class="visit-type__select-visit-type visit-type__select-service-type--disabled" name="visitType" data-model="visitType" data-feedback="visitType" id="visit-type" aria-label="Select Visit Type" disabled="">
          
[...]

PR #183, opened 2016-01-14
Fix for #176 - stickit unbind correctly

mandragorn:
Correctly unbinding stickit on undelegate so that when it is bound during delegate it does not screw up the DOM.

### DIFF
--- a/modules/FormView.js
+++ b/modules/FormView.js
@@ -119,6 +119,17 @@
     },
 
     /**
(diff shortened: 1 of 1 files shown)


mandragorn:
:-1: tests are still failing, but fewer.  Seems like stickit keeps state and removing it looses that state.  So any bindings setup before an unstick (undelegate) are lost when the view re-rendered.

PR #188, opened 2016-01-28
Fix #185, adds unit test reproducing problem

### LINKED ISSUE #185: Error when initializing Form View
I'm getting an error when initializing a form view. This error is caused because FormView.constructor calls InitializedFormView.initialize before this.fields has been initialized. InitializedFormView.render calls FormView.delegateEvents() which eventually calls FormView.__getFieldOptions().

There are several options I see for resolving this bug:

1. Move the field initialization to above View.apply
2. Modify __getFieldOptions such that it runs even if this.fields isn't initialized.

What is the justification for having the call to View.apply(this, arguments) be in the middle of FormView.constructor?

Earthstar:
Lets try this again


kentmw:
+2

Issue #192, opened 2016-02-08
Switch test browser environment from jsdom to Karma + PhantomJS 


Earthstar:
The current test infrastructure with jsdom has some deficiencies:

- Doesn't run on windows
- Test setup isn't as straightforward as it could be
- Can only test in NodeJS environment, not actual browsers

Running the tests with Karma would eliminate these problems.

arikwex:
Should #113 be closed in favor of this?

kentmw:
+1 to this issue and lets close #113

PR #193, opened 2016-02-09
Work on #192

Earthstar:
My strategy for issue #192 is to add Karma as a test runner, slowly port over tests, and then delete the JSDom setup once we have equivalent coverage.


kentmw:
Can I just say that im super excited for this to go through?

Earthstar:
@kentmw If someone gives me a +2, it can be merged imminently :D

kentmw:
I like the method you took to have both run at the same time. Although updates to existing tests will have to be duplicated in both the karma and jsDom setups. I'd suggest if you port over a test that you remove the jsDom ones so that we don't have to maintain both. Can you finish porting over the view attach spec and delete the jsdom one? Also make a depricated note on the jsdom tests (maybe a top of file) that suggests the file is migrated to karma before updates happen.

On the note about spying in the view. I'm not sure I'm a fan of having to attach the spies on each test spec just because we could reuse the views outside of tests (which they won't). We can keep it now but if it gets too repetitive, we might want to move it back into the source file.

Also we'll have eventually update the testing gulp files so that we can run singular test specs, etc.

Lastly, are you planning on having specs live alongside the source code, instead of separating them into different directories? What's the motivation for that?

Earthstar:
> Can you finish porting over the view attach spec and delete the jsdom one?

I can certainly delete the old specs as they are ported over. However:
1. I only have 4 hours of CS to spend on torso each week. Therefore, I'd like to merge things in small increments. I think the value provided by getting the Karma infrastructure merged outweighs the messiness of having a partially-ported test. Once Karma is set up, other people can help me port over tests \*wink wink nudge nudge*
2. I want to improve the test descriptions as I'm porting them over, but I don't know what the intent is behind some of the tests. Saying something was done "correctly" doesn't tell me what the desired behavior is. I'd need some help describing them.

> On the note about spying in the view ...

I played around with having a view generator like in the previous tests. The problem is that you can't create a Jasmine spy outside of a test block. I find it awkward to have to require and instantiate the view from inside a describe block as opposed to just importing the view.
I prefer using a hook as opposed to defining the spies in the view class because then you can identify which test is using which spies.

> Are you planning on having specs live alongside the source code, instead of separating them into different directories?

Are you talking about the fact that the test views and the specs live in the same directory? I can split test classes into their own directory.

kentmw:
I'm all for doing things in small chunks. I also understand that people come and go without completing new features.

Side note: 4 hours a week is what you get _paid_ to do. I don't get paid to work on Torso, unless being contracted by Vecna - it's an open source project. You can spend some free time on it if you care about it :)

Porting viewAttach to your viewAttachSpec should have been as easy as copying all of the "it" blocks. With your change to the spies, it requires making updates to the source files as well, I understand that. Either finish viewAttach port or at least finish the it('can be initalized correctly') and it('can be attached correctly') blocks and delete those blocks.

Once this gets merged, don't worry, I will help move test files into Karma because I think it's fun and useful.

Yes, please move test source files into its own directory.

It's unclear to me if the way you are defining dependencies (like underscore or jquery) in your test source files are actually getting the dependencies from the window object (or the phantomjs instance) or just grabbing them from node_modules. If it's from the node_modules, this won't work and you need a generator as soon as you need something DOM-depenedent like jquery. If it is pulling from the dom environment, then that's great and we don't need generators. I still would rather put the spies on the class level instead of at the test level. Your way right now has lots of boilerplate both in the test code and has to manipulate the acutal source classes with spy hooks - something that should be hidden to the code being tested.

Earthstar:
Re: defining dependencies. The way the Karma setup works is that Karma uses each test file as the root for browserify to pull dependencies. The browserified test file is loaded into the browser where it executes. Because of this, in the test file, importing modules is as simple as calling "require".

test/karma/helpers/ParentClickView.js
```diff
@@ -0,0 +1,35 @@
...
+module.exports = TorsoView.extend({
+  events: {
+    'click div.parent' : 'myClick'
+  },
+  initialize: function() {
+    this.spyHook();
+
+    spyOn(this, 'myClick');
+    spyOn(this, 'afterMyEvent');
+    spyOn(this, 'afterMyDeactivatableEvent');
```
kentmw: okay, i did some thinking about spyOn's. I don't like the testing framework to be in our source objects. Source files, sure, but someone should be able to look at the Torso View or Model and not see any manipulation for the testing framework. This is because we don't want them confused about what's necessary for the functionality vs the testing.

So, if we care about this, manipulating the initialize method whether its to setup spys or a hook for setting up spys is not ideal.

We have two options to handle this:
1. every time we instantiate the view in the test, we add the spy on the initialized view. Most cases, we don't apply any spies because we don't use them very often. I'm acutally okay with this if you think this is the best option.
2. Use the backboneSpy.js that was already created for this exact reason, that adds a prototype-level wrapper that finds the instance that was invoking it and adds the spy. All jsdom tests use the spyOnBackbone thing, which makes spies work on the prototype.

Earthstar: It sounds like to me, the reason you don't want spyOn to appear in the View's initialize is because people would look at the test classes as an example for how to instantiate real Torso views. I tried to only use spyHook to attach listeners, but it got messy with ParentClickView because the ChildClickViews are being instantiated inside ParentClickView.initialize. Would it be acceptable to you if we indicated that the views are instrumented for testing by naming them Spied*View.js?

kentmw: That's not really fixing it - rather than to separate test source objects and test framework code, that's just further mixing it by naming the test source object by a seemingly insignificant test framework attachment.

Any reason you don't want to do one of the two options I suggested?

kentmw:
duplicate

Issue #123, opened 2015-09-09
Has anyone figured out how to animate a transition from one template render to another?


kentmw:
If I produce one set of html to another across two renders and the second render takes up more space, let's say, I want to know how to animate that transition. As in, I'd like the container div to expand.

This probably could be solved many different ways, potentially not using render at all, but I was curious if others have wanted/tried this before.

kentmw:
As of now, I'm just rendering all new content with hidden classes, then animating them in after render is complete

arikwex:
This is considered a standard approach.
If you look at angularJS for the ngAnimate support, you'll see that they do the same thing.
There is lots of voodoo magic in angular, but it pretty much listens for when CSS animations are done before removing the old element.  When a transition happens there is a sequence of CSS class triggers:
1) .ng-enter is placed on the newly entering DOM
2) .ng-leave is placed on the current DOM
3) .ng-enter.ng-enter-active is placed on the new DOM
4) .ng-leave.ng-leave-active is place on the current DOM

This should handle your animation start and stop states.  The hard part is definitely just how to hook in to the "When is the CSS animation done."

kentmw:
The new transitions code allows for all of the css class additions (they don't happen by default, but are if we find that all transitions follow this same behavior, we can roll that in by default). It also allows for custom javascript play as well.

Currently, we're letting the developers dictate "when the css animation is done". Matt's looking into how to track css animation completions generically.

PR #194, opened 2016-02-10
Transition logic #123

kentmw:


kentmw:
@catbieber @mandragorn @arikwex There are two parts to this commit. The main part is the transition logic with an addition to the API:

  transitionIn - defines how a view can transition/animate into the dom
  transitionOut - defines how a view can transition/animate out of the dom
  transitionInView - call from a parent view to transition in a tracked (child) view.
  transitionSiteToNewView - an alternative to injectView which will find a view already in an injection site and transition it out and transition in the new view at that site.

The second part is an update to the render method. The high level change here is that the render method is no longer meant to be overridden, instead opting to connect common render logic (like injecting views) through hooks. There is a part that has a breaking change: changing the name "plug" and "unplug" to "prerender" and "postrender". I also added some triggering events that may be useful  ('render-begin', 'render-before-dom-replacement', 'render-after-dom-replacement', 'render-complete'). Here is the flow of the render method:

1. trigger 'render-begin'
2. invoke prerender
3. trigger 'render-before-dom-replacement'
4. detach all tracked views
5. template rerender (hotswap)
6. delegate events
7. trigger 'render-after-dom-replacement'
8. invoke attachTrackedViews (this hook is where you will inject and transition in your child/tracked views)
9. invoke postrender
10. trigger 'render-complete'

Now, I understand that while this change in the render method makes life WAY more convenient to write views and also is necessary when determining which views were previously at an injection site, this it is a major change, one that seems tangential to the core of the transition logic. We can talk about that more together.

kentmw:
Also there's a demo project: https://github.com/kentmw/torso-transition-demo. You'll need to checkout the transition branch of torso and npm link it. Then npm link the demo project's backbone-torso dependency.

kentmw:
Self notes/peer review:

1. Update documentation to reflect new changes (esp. for render)
2. Add in-torso tests

kentmw:
First pass at documentation and test cases done.

modules/View.js
```diff
@@ -154,32 +158,130 @@
...
+        return this.injectView(injectionSiteName, newView, options);
+      }
+      if (!previousView) {
+        // Only transition in the new current view and find the injection point.
+        injectionPoint = this.$('[inject=' + injectionSiteName + ']');
+        return this.transitionInView(injectionPoint, newView, options);
+      }
+      this.injectView(injectionSiteName, previousView, options);
+      options.cachedInjectionSite = previousView.injectionSite;
+      newInjectionSite = options.newInjectionSite = $('<div inject="' + injectionSiteName + '">');
```
mandragorn: should the injection site be a span so its not a block element?

Also I'm not quite following what the newInjectionSite is used for.

modules/View.js
```diff
@@ -698,22 +721,79 @@
...
     /**
+     * Gets the hash from id to views of the correct views given the options.
+     * @param [options={}] {Object}  Optional options.
+     *   @param [options.shared=false] {Boolean} The view is a shared view instead of a child view
+     *                                           (shared views are not disposed when the parent is disposed)
+     * @method __getTrackedViewsHash
+     * @return the object map cotaining either the shared or child views keyed by their cid
+     * @private
+     */
+    __getTrackedViewsHash: function(options) {
```
mandragorn: should this work the same as getTrackedViews() with an option for shared an child separately (and no options returns both)?

modules/View.js
```diff
@@ -698,22 +721,79 @@
...
+     * Used internally by Torso.View to keep a cache of tracked views and their current injection sites before detaching during render logic.
+     * @private
+     * @method __updateInjectionSiteMap
+     */
+    __updateInjectionSiteMap: function() {
+      var parentView = this;
+      this.__injectionSiteMap = {};
+      _.each(this.getTrackedViews(), function(view) {
+        if (view.isAttachedToParent() && view.injectionSite) {
+          parentView.__injectionSiteMap[view.injectionSite.attr('inject')] = view;
```
mandragorn: are we guaranteed the injection site is defined by Torso?  If not then people can inject views based on selectors other than inject.

torso-bundle.js
```diff
@@ -2046,22 +2130,79 @@
...
+    __getLastTrackedViewAtInjectionSite: function(injectionSiteName) {
+      // check to see if a view was cached before a render
+      var previousView = this.__injectionSiteMap[injectionSiteName];
+      if (previousView) {
+        // make sure previous view is still tracked
+        previousView = _.contains(this.getTrackedViews(), previousView) ? previousView : undefined;
+      } else {
+        // if not, see if a view is currently in the injection site.
+        var matchingViews = this.getTrackedViews().filter(function(view) {
+          return view.injectionSite && view.injectionSite.attr('inject') == injectionSiteName && (!excludingViews || !_.contains(excludingViews, view));
```
mandragorn: where is excludingViews defined?

modules/View.js
```diff
@@ -698,22 +721,79 @@
...
+    __getLastTrackedViewAtInjectionSite: function(injectionSiteName) {
+      // check to see if a view was cached before a render
+      var previousView = this.__injectionSiteMap[injectionSiteName];
+      if (previousView) {
+        // make sure previous view is still tracked
+        previousView = _.contains(this.getTrackedViews(), previousView) ? previousView : undefined;
+      } else {
+        // if not, see if a view is currently in the injection site.
+        var matchingViews = this.getTrackedViews().filter(function(view) {
+          return view.injectionSite && view.injectionSite.attr('inject') == injectionSiteName && (!excludingViews || !_.contains(excludingViews, view));
```
mandragorn: where is excluding views defined?

test/source/transitionUsingParentViewGenerator.js
```diff
@@ -0,0 +1,40 @@
+var spyOnBackbone = require('../spec/functional/backboneSpy'),
```
mandragorn: do you need to do anything with spyOnBackbone after requiring it?  I don't see it used below.  Same for most of the view files.

test/spec/functional/viewTransition.js
```diff
@@ -0,0 +1,59 @@
+var testSrcPath = '../../source',
+    spyOnBackbone = require('./backboneSpy');
+
+describe('A View that can transition in tracked views', function() {
+
+  var env, _, $, ParentView;
+
+  beforeEach(function(done) {
+    require('./clientEnv')().done(function(environment) {
+      env = environment;
```
mandragorn: local vars using this is best practice with Jasmine (as far as I know).

this.env = environment;
this.$ = ...
...

Then you can access the vars in your it and describe blocks with this.env, etc.  This prevents any leaking state from crossing tests.

test/spec/functional/viewTransition.js
```diff
@@ -0,0 +1,59 @@
...
+      $('body').append('<div class="app"></div>');
+      done();
+    });
+  });
+
+  it('can transition in the first view into an injection site', function(done) {
+    var parentView = new ParentView();
+    parentView.render();
+    expect(parentView.$el).toBeDefined();
+    parentView.transitionPromise.done(function() {
```
mandragorn: should render return a promise that resolves when all descendents have completed rendering?

modules/View.js
kentmw: We could do that (I actually finished doing this change) and it cleaned up getTrackedViews method. But, it also causes a potential issue: "returning" both hashes can be done by merging the two hashes into a new one and returning it - great for read-only manipulations. But registerTrackedView expects to be able to change (add to) the hash that is returned, so it must make sure that it's passing in an option.

Do we want to have this read/write - read-only mix of returns or just keep it as it because it's different - it deals with the hashes directly which can only be one or the other.

modules/View.js
kentmw: Yes, injecting views is done through injectView and is always done by Torso

modules/View.js
kentmw: old logic. removed

test/source/transitionUsingParentViewGenerator.js
kentmw: removed

modules/View.js
mandragorn: why not write specific methods for operations and assume that you shouldn't manipulate read-only responses.

i.e. addTrackedView() or removeTrackedView()

modules/View.js
mandragorn: ahh, didn't realize this was a hard requirement now (using inject attribute).

kentmw: Hmm... this made me re-look into it. this.injectView relies on defining "inject" attr. this.attach which is used by this.attachView does not rely on it.

If people use attachView and not injectView, which I guess is possible even if that's not what I do, then the "inject" attr will not be defined.

The outcome of this use case is that if you use attachView and don't use a standard injection site, and then call transitionNewViewIntoSite with an injection site name and want/hope that it will automatically find the old view somehow to transition it out, that view will not be found. Which makes sense that it wouldn't find it, because come on, transitionNewViewIntoSite is looking for the same transition site, and you didn't even give one to the first child view.

And on top of all that, if you did want to transition out a view that wasn't given a "proper" injection site, you can always specifiy it as the option "previousView" when calling transitionNewViewIntoSite.

I think we're okay

mandragorn: ok, as long as the docs are clear.

modules/View.js
kentmw: We can definietly make it a span.

Here's the process:
We re-render like normally, injecting in the old view, then adding an equivolent injection site next to the old view, finally transitioning in the new view into the new injection site.

The reason I did this instead of just plopping the new view in "before" or "after" the old view without an injection site, is because this way we get to reuse the same logic/methods to transition in the new view (using an injection site). Plus, the new view needs a stored injection site anyway (so that it can be transitioned out later). It made sense to follow the same pattern.

modules/View.js
kentmw: i'll make __getTrackedViewsHash always read-only.

addTrackedView and removeTrackedView is effectively the same thing as register and unregister tracked view. I'll just have them do the logic manually.

modules/View.js
```diff
@@ -86,12 +87,26 @@
      * @method render
      */
     render: function() {
-      this.unplug();
```
pepperbc: This and ListView seem to have a lot of common code - can we reuse that somehow?

test/spec/functional/viewTransition.js
kentmw: great question - one that I considered. I didn't want to complicate the process.

But, I gave it a go. prerender, postrender, and attachTrackedViews all can return a promise (or promises) if they want or not and render will return one promise that is resolved when all of them are resolved.

modules/View.js
```diff
@@ -116,12 +131,6 @@
      * See Torso.templateRenderer#render for params
      */
     templateRender: function(el, template, context, opts) {
-      // Detach just this view's child views for a more effective hotswap.
-      // The child views will be reattached by the render method.
-      this.detachTrackedViews({ shared: false });
-      // Detach just this view's shared views for a more effective hotswap.
-      // The shared views will be reattached by the render method.
-      this.detachTrackedViews({ shared: true });
       templateRenderer.render(el, template, context, opts);
```
pepperbc: This is just a pass-through now - can we remove it?

test/spec/functional/viewTransition.js
kentmw: done. can't say i like how it ended up looking - most of what i was doing was in nested functions with different this contexts.

modules/View.js
kentmw: good question and valid point. I've had to go back a few times and update the list view to mirror a normal view. Not easy to refactor but I can look into it.

modules/View.js
kentmw: Two reasons I want to keep it:
1. backwards compatibility. not such a huge reason but still small.
2. maintaining a separate interface. templateRenderer is actually a different module - one that we have a high likelihood of replacing in the future (with something like the new handlebars library, Bars). The implementation of our templateRender may change but our method hopefully will not.

modules/View.js
```diff
@@ -154,32 +158,130 @@
...
+     *   newView: the new view
+     *   previousView: the previous view (can be undefined)
+     *   parentView: the parent view transitioning in or out the tracked view
+     * }
      */
-    injectView: function(injectionSite, view, options) {
-      var injectionPoint = this.$('[inject=' + injectionSite + ']');
-      if (view && injectionPoint.size() > 0) {
-        this.attachView(injectionPoint, view, options);
+    transitionNewViewIntoSite: function(injectionSiteName, newView, options) {
```
pepperbc: When would I use this instead of injectView with the "useTransistion" option.

modules/View.js
```diff
@@ -317,38 +420,57 @@
...
 
     /**
-     * Gets the hash from id to views of the correct views given the options.
-     * @param [options={}] {Object}  Optional options.
-     *   @param [options.shared=false] {Boolean} The view is a shared view instead of a child view
-     *                                           (shared views are not disposed when the parent is disposed)
-     * @method __getTrackedViewsHash
+     * Override to provide your own transition out logic. Default logic is to just detach from the page.
+     * @method transitionOut
+     * @param done {Function} callback to be invoked when the transitions is complete *MUST BE CALLED*
```
pepperbc: "Must be present" instead of "Must be called" ?

modules/View.js
kentmw: no difference. up for hearning how to handle it differently. injectView is go-to for attaching child views and I wanted transitions an option in that method. I also wanted to keep this block of logic as a unit. Only thing I can think of is making this method private?

modules/View.js
```diff
@@ -459,159 +545,101 @@
...
      * @return {View} the tracked view
      * @method unregisterTrackedView
      */
-    unregisterTrackedView: function(view, options) {
-      var trackedViewsHash = this.__getTrackedViewsHash(options);
-      delete trackedViewsHash[view.cid];
+    unregisterTrackedView: function(view) {
+      delete this.__childViews[view.cid];
+      delete this.__sharedViews[view.cid];
       return view;
```
pepperbc: Why are childViews and sharedViews in separate hashes?

modules/View.js
kentmw: the callback "done" must be invoked at sometime during the transitionOut method. Very much how Jasmine tests must invoke done()

modules/View.js
kentmw: i believe its mostly a trade off decision because the map is cid to views and adding another parameter would make it bulky
[2/16/16, 1:16 PM] Kent Willis (kmwillis@mit.edu): This was the alternative i thought of, and I didn't like it - especially concerning backwards compatabiltiy.
{
  cid: {
     view: myView,
     shared: true
  }
}

modules/View.js
```diff
@@ -372,63 +493,39 @@
...
-    },
-
-    /**
-     * @return all of the shared views this list view has registered
-     * @param [options={}] {Object}  Optional options.
-     *   @param [options.shared=false] {Boolean} The view is a shared view instead of a child view
-     *                                           (shared views are not disposed when the parent is disposed)
+     * Returns all tracked views, both child views and shared views.
      * @method getTrackedViews
+     * @param [options={}] {Object}  Optional options.
```
pepperbc: If both are true, should we return everything? Intuitively I would expect that to happen, and it would also leave us in a better place in cases a third type is ever added. I know the docs say "only" and don't define that behavior, but arbitrarily returning one of them is unexpected.

modules/View.js
```diff
@@ -698,22 +721,79 @@
...
         this._detached();
         this.__attachedCallbackInvoked = false;
       }
-      _.each(this.__childViews, function(view) {
-        view.invokeDetached();
-      });
-      _.each(this.__sharedViews, function(view) {
-        view.invokeDetached();
+      _.each(this.getTrackedViews(), function(view) {
+        // No need to check if tracked view is actually detached from the DOM as long as they are attached to the parent and the if parent is detached,
```
pepperbc: Unsure what this comment means

modules/View.js
```diff
@@ -459,159 +545,101 @@
...
-     *   @param [options.shared=false] {Boolean} The view is a shared view instead of a child view
-     *                                           (shared views are not disposed when the parent is disposed)
+     *   @param [options.shared=false] {Boolean} If true, registers view as a shared view. These are views not owned by this parent. As compared to a child view
+     *                                           which are disposed when the parent is disposed. If false, registers view as a child view which are disposed when the parent is disposed.
      * @return {View} the tracked view
      * @method registerTrackedView
      */
     registerTrackedView: function(view, options) {
+      options = options || {};
+      options.child = options.shared ? false : true;
```
pepperbc: Using this, I can register a shared view and a child view using the same cid - this would probably lead to weird behavior. Could happen theoretically if you're registering in two places in your code and forget to put "shared" in one of them

modules/View.js
pepperbc: "MUST BE CALLED" wasn't clear to me what this meant until I reread it a couple times. Maybe in the general description as well say something as well?

modules/View.js
kentmw: Would calling unregisterTrackedView at the top of the method help. That will remove it from both/either before adding it again.

modules/View.js
```diff
@@ -459,159 +545,101 @@
...
-     */
-    activateChildViews: function() {
-      this.activateTrackedViews({ shared: false });
-    },
-
-    /**
      * Binds the view as a tracked view - any recursive calls like activate, deactivate, or dispose will
-     * be done to the tracked view as well.  Except dispose for shared views.
-     *
+     * be done to the tracked view as well.  Except dispose for shared views. This method defaults to register the
```
pepperbc: Minor organizational comment - I prefer add/remove/change methods to be right next to each other so when reading you can see the entire lifecycle at once. If this is organized some other way thats fine as well.

modules/View.js
kentmw: done

modules/View.js
kentmw: sounds good.

modules/View.js
pepperbc: If transitionNewViewIntoSite was an analog of attachView instead of injectView, and injectView still did all the extra logic, this confusion could be avoided. I feel like we should either fully support attachView (and store attached views indexed by dom element), or not at all. Having two similar ways of attaching views with different features and internal workings seems like a mess

test/source/parentClickViewGenerator.js
```diff
@@ -16,8 +16,8 @@ module.exports = function(window) {
     events: {
       'click div.parent' : 'myClick'
     },
-    render: function() {
-      this.$el.html("<div class='parent'>test</div><div inject='one'></div><div inject='two'></div>");
+    template: "<div class='parent'>test</div><div inject='one'></div><div inject='two'></div>",
+    attachTrackedViews: function() {
```
pepperbc: The dissonance between having attachTrackedViews use injectView instead of attachView is confusing - will create and additional issue regarding naming

torso-bundle.js
```diff
@@ -305,6 +305,51 @@
           },
 
           /**
+           * Will force the cache to fetch a subset of this collection's tracked ids
+           * @method requesterMixin.fetchSubsetOfTrackedIds
```
pepperbc: Is this related?

torso-bundle.js
```diff
@@ -305,6 +305,51 @@
...
+           * @return {Promise} promise that will resolve when the fetch is complete
+           */
+          fetchSubsetOfTrackedIds: function(ids) {
+            var subsetOfTrackedIds = _.intersection(ids, this.getTrackedIds());
```
pepperbc: Is there a use case where you wouldn't want to track the missing ids? Even if so, is it common enough that we wouldn't want to default to tracking anyhow?

modules/View.js
kentmw: hard to know what the right order is. There are a lot of tracked view get/has/change methods. They are mostly bundled together already but again there are many of them. Also there's the organization of flow of rendering, or how common the method is.

modules/View.js
kentmw: done

modules/View.js
kentmw: done

modules/View.js
pepperbc: Since there's no difference in functionality I'd prefer this method to be private. It saves migration headaches later, and makes the API less complicated

modules/View.js
pepperbc: Agree - can't think of an elegant solution, and not that big a priority for now.

modules/View.js
pepperbc: That's what I was thinking in the back of my head, sounds good.

modules/View.js
kentmw: done

modules/View.js
kentmw: I gave a go at reorganizing. Made a number of methods private that were rarely used by developers and mostly only used by Torso methods. Then I grouped methods by topic: all attach-related methods near each other, all detach-related methods near each other, etc.

test/source/parentClickViewGenerator.js
kentmw: agreed. removed injectView and made attachView handle both attachView logic and  injectView logic.

modules/View.js
kentmw: I'd like to open another issue to handle this. I'd like this to be handled separately from this commit if possible.

modules/ListView.js
```diff
@@ -295,7 +300,11 @@
           // Shouldn't get here. Item view is missing...
         }
       }, this);
+      renderPromises.push(this.attachTrackedViews() || $.Deferred().resolve().promise());
+      this.__injectionSiteMap = {};
+      renderPromises.push(this.postrender() || $.Deferred().resolve().promise());
       this.trigger('render-complete');
+      return $.when.apply($, _.flatten(renderPromises));
```
mandragorn: why can't you call when directly?  $.when(_.flatten(renderPromises));

modules/View.js
```diff
@@ -83,32 +100,54 @@
...
      * function that accepted a single JSON context.
+     * The render method returns a promise that resolves when rendering is complete. Typically render
+     * is synchronous and the rendering is complete upon completion of the method. However, when utilizing
+     * transitions/animations, the render process can be asynchronous and the promise is useful to know when it has finished.
      * @method render
+     * @return {Promise} a promise that when resolved signifies that the rendering process is complete.
      */
     render: function() {
-      this.unplug();
+      var renderPromises = [];
```
mandragorn: thoughts on a single render method in view with each state as a private method.  ListView can override the private methods for custom logic.

modules/View.js
```diff
@@ -154,81 +182,204 @@
...
+     * @method attach
+     */
+    attach: function($el, options) {
+      options = options || {};
+      var injectionSite,
+        replaceMethod = options.replaceMethod,
+        discardInjectionSite = options.discardInjectionSite;
+      if (!this.isAttachedToParent()) {
+        this.render();
+        this.injectionSite = replaceMethod ? replaceMethod(this.$el) : $el.replaceWith(this.$el);
```
mandragorn: for readablility:
```
if (...) {
  this.render();
  this.__replaceInjectionSite(_.pick(options, 'replaceMethod', 'discardInjectionSite'));
  this.__cleanupAfterReplacingInjectionSite();
}
```

modules/View.js
```diff
@@ -671,47 +589,209 @@
...
+     * @return {Promise} resolved when all transitions are complete. No payload is provided upon resolution.
+     * When the transitionIn and transitionOut methods are invoked on the new and previous views, the options parameter will be passed on to them. Other fields
+     * will be added to the options parameter to allow better handling of the transitions. These include:
+     * {
+     *   newView: the new view
+     *   previousView: the previous view (can be undefined)
+     *   parentView: the parent view transitioning in or out the tracked view
+     * }
+     */
+    __transitionNewViewIntoSite: function(injectionSiteName, newView, options) {
```
mandragorn: this method feels long, does it make sense to have private methods for readability and make it clear what the major steps of the method are?

modules/ListView.js
kentmw: $.when takes a comma separately set of arguments. I have an array. I use apply to convert an array into separate arguments

modules/View.js
```diff
@@ -671,47 +589,209 @@
...
+      });
+      newView.transitionIn(function() {
+        parentView.attachView($el, newView, options);
+      }, currentDeferred.resolve, options);
+      return currentDeferred.promise();
+    },
+
+    /**
+     * Gets the hash from id to tracked views. You can limit the subset of views returned based on the options passed in.
+     * NOTE: returns READ-ONLY snapshots. Updates to the returned object will not be saved nor will updates to the underlying maps be reflected later in returned objects.
```
mandragorn: _.extend() is a shallow copy, right?  if you change a property of the object it will be reflected (probably desirable since you can then interact with the view).  So it is only a snapshot of the cid -> view mapping.

modules/View.js
kentmw: done

modules/View.js
kentmw: done

modules/View.js
kentmw: yes, i will make that clear

modules/View.js
kentmw: done

modules/ListView.js
```diff
@@ -283,9 +286,17 @@
       } else if (this.emptyTemplate) {
         injectionSite.replaceWith(this.emptyTemplate(this.prepareEmpty()));
       }
-      this.trigger('render-before-dom-replacement', newDOM);
       this.$el.html(newDOM.contents());
-      this.delegateEvents();
+    },
```
mandragorn: maybe rename to __addDOMTo$el() or something.  It isn't clear from the method name that it updates this.$el.  I was expecting it to return the dom fragment (but it makes sense to include the add since it is different between list and view).

modules/ListView.js
```diff
@@ -295,7 +306,6 @@
           // Shouldn't get here. Item view is missing...
```
mandragorn: Should this throw an error so the developer knows what went wrong (that something went wrong)?

modules/ListView.js
kentmw: how about __updateViewsDOM

torso-bundle.js
```diff
@@ -1431,32 +1509,44 @@
...
      */
     render: function() {
-      this.unplug();
-      if (this.template) {
-        this.templateRender(this.$el, this.template, this.prepare());
-      }
-      this.plug();
+      var renderPromises = [];
+      this.trigger('render-begin');
+      addPromises(renderPromises, this.prerender());
```
mandragorn: I could see having addPromises be inline (it is one of those things that is easier read at this level).  Also I could convince myself that this is more readable assuming I can guess what addPromises does.

torso-bundle.js
```diff
@@ -1431,32 +1509,44 @@
...
+      this.trigger('render-before-dom-replacement');
+      this.__createDOM();
```
mandragorn: yea, this should be clear that it is putting the dom somewhere after creating it, not just returning it.

modules/ListView.js
kentmw: I think that may be unnecessarily jarring. For whatever reason a view not existing to a model in modelsToRender is not necessarily a programmer error - so I'd rather not throw a thread-ending error.

torso-bundle.js
```diff
@@ -1431,32 +1509,44 @@
...
       this.delegateEvents();
+      this.trigger('render-after-dom-replacement');
+      addPromises(renderPromises, this.attachTrackedViews());
+      this.__injectionSiteMap = {};
+      addPromises(renderPromises, this.postrender());
+      this.trigger('render-complete');
+      return $.when.apply($, _.flatten(renderPromises));
```
mandragorn: I like it.  Really clean for a high-level function.

torso-bundle.js
```diff
@@ -1431,32 +1509,44 @@
...
-      this.plug();
+      var renderPromises = [];
+      this.trigger('render-begin');
+      addPromises(renderPromises, this.prerender());
+      this.trigger('render-before-dom-replacement');
+      this.__createDOM();
       this.delegateEvents();
+      this.trigger('render-after-dom-replacement');
+      addPromises(renderPromises, this.attachTrackedViews());
+      this.__injectionSiteMap = {};
```
mandragorn: we clear the injection site map here?  That seems weird to me to clear it after attaching all the views and swapping out the injection sites.  I was expecting it to be a map that persists between rendering so you can revert the injection sites.

torso-bundle.js
```diff
@@ -1786,256 +1872,404 @@
...
-    activateChildViews: function() {
-      this.activateTrackedViews({ shared: false });
+    __createDOM: function() {
+      if (this.template) {
+        this.__updateInjectionSiteMap();
+        // Detach this view's tracked views for a more effective hotswap.
+        // The child views should be reattached by the attachTrackedViews method.
+        this.detachTrackedViews();
+        if (_.isString(this.template)) {
+          this.$el.html(this.template);
```
mandragorn: so if you use strings it doesn't use template renderer

torso-bundle.js
kentmw: This line is unnecessary and I will remove it.

On that note, though, I only need to have the map be up-to-date right before render - which happens in __updateDOM

torso-bundle.js
kentmw: made it always use template renderer

modules/ListView.js
mandragorn: that works.  the update prefix is clearer that something will be changed.

modules/ListView.js
mandragorn: ok

kentmw:
@pepperbc said we could merge it and he'll look over it with a finer comb come next week.

Issue #207, opened 2016-02-22
Does new render logic need to handle multiple templates?


kentmw:
Some projects are using multiple templates to define different components of their DOM. Should we support this? Should we suggest multiple views in this case? or use partials?

mandragorn:
I've been having a lot of cases recently where I have a view with all the logic I need, but I want to style it slightly differently (i.e. for an admin user vs. patient) which causes me to create a lot of css class switches and in some cases different DOM elements.

Partials have been problematic in the past and there is something about browserify not fully supporting them (they are not processed correctly by quick-sip).

I'm not sure if multiple views or templates are better.  But for generality multiple templates are probably good to support.  Would it make sense to have a template context property reserved for choosing the template so that the logic for generating the template context and determining the template to use are all in one place (prepare())?  We could default the property to the view's template to reserve the use of the property.

kentmw:
@mandragorn added a pull request for this. I think it just makes sense to allow people to put in whatever dom update logic they want.

kentmw:
open ```updateDOM``` to be overridable,

PR #208, opened 2016-02-22
#207 - allowing custom DOM update logic in the render process

kentmw:


docs/DOCUMENTATION.md
```diff
@@ -120,6 +120,8 @@ Events are also triggered at different points in the render method:
 
 It is rare, and discouraged to override the render method. Instead, use the available hooks to control the render logic.
 
+The ```updateDOM``` method is called during step 3 during DOM creation and by default, invokes: ```this.templateRender(this.$el, this.template, this.prepare());```. If you wish to have custom logic, override this method.
+
```
mandragorn: make a note that listView does more complicated things so be careful when overriding a from ListView?

mandragorn:
+1

Issue #210, opened 2016-02-23
Redefining a FormView's FormModel in the initialize method needs event listening cleanup


kentmw:
You can't redefine a FormView's FormModel in the initialize method without also redefining the this.listenTo(this.model, ...) blocks.

It seems like it would be better to define the this.listenTo(this.model ...) blocks after the View.apply - that way you can be sure the model is the correct form model before attaching listeners to it.

mandragorn:
+1

Issue #209, opened 2016-02-22
addComputed() syntax could be more elegant


pepperbc:
The addComputed() syntax is quite verbose for simple cases, and is a bit confusing even with the documentation. We should both fix up the documentation and think about changing the API to be cleaner.

kentmw:
For inspiration (and history of what was used as a basis for current implementation), check out https://github.com/NYTimes/backbone.stickit

kentmw:
Okay, so another issue is that defining the mappings require the actual object models which is hard/impossible to do from the class level.

Right now, you can define the mappings by passing them in as arguments during initialization (because assumingly when you create the form model, you have access to the object models) or you call addModel and addComputed later. There is a little known third option called "mapping" which is actually mislabeled as "defaultMapping" in the docs. You can set this but only if you have access to the object model when you are doing your class definition - which is like never.

Here's my proposed solution to both not being able to define mappings on a class level and hopefully cleaning up the syntax: rely on named object models on the form model. Here's what it would look like:

```
var MyFormModel = Torso.FormModel.extend({
    mappings: {
      myModel: ['username', 'email'],
      myOtherModel: true, // grabs all fields
      myThirdModel: 'password',
      myFourthModel: ['name', 'ssn'],
    },

    computeds: [{
      myModel: ['dob', 'ssn'],
      myThirdModel: 'password',
      pull: function(fields) {
        // fields.myModel.dob
        // fields.myModel.ssn
        // fields.myThirdModel.password
      },
      push: function(models) {
        // models.myModel
        // models.myThirdModel
      }
    }, {
      myModel: 'dob',
      pull: function(fields) {
        // fields.myModel.dob
      },
      push: function(models) {
        // models.myModel
      }
    }],

   initialize: function(options) {
     this.myModel = /* find a model */
     this.myOtherModel = options.myOtherModel || /* create one*/
     this.myThirdModel = options.myThirdModel || /* create one*/
     this.myFourthModel = /* find my fourth model */
  }
});
```

Couple of notes both good and bad on it:
- The FormModel constructor will run the mapping initialization after the initialize method is done.

- Once the initialization happens, the objects will be stored in separate configs so switching out the objects later ( ```this.myModel = aNewModel``` ) won't have any affect.

[...]

Issue #70, opened 2015-06-18
When using a FormModel to create new models, sometimes it doesn't push all values


kentmw:
I believe that it will push all the FormModel values to the model if the model contains NO values at all. But, if there are any defaults in the model (like @Class attributes), then it will default to just those properties.

Perhaps we can rework the logic so FormModel's work fine pushing to new object models without having to specify the fields it tracks.

kentmw:
This is because of this line in __copyFields:
```
      if (!fields && this === origin) {
        fields = _.keys(destination.attributes);
      }
```
which assumes that if you leave the fields blank, and you are performing a push, you want to copy only the properties available in the destination which is the object model. In other words, a push will only push properties that exist on the object model. For pulls, it will pull all the object model values.

If we change this, or remove this conditional, we run the risk of pushing other object models' fields to the one tracking all the attributes. I.e. if you track all the fields on Object Model A and only a few in Object Model B, and you invoke a push, which fields are pushed to Object Model A? All of them? Including Object Model B? Including all the computed values? This is why we are currently trying to grab which attributes are Object Model A's by filtering on fields that are currently present on ObjectModel A. Which, of course, doesn't work if ObjectModel A is new and we're using the FormModel to create it.

Which leaves us with a couple of approaches:

1. Find out if a model is new and push all the properties if it is
2. Don't support unexplicit "all" properties when mutliple models are used

or I think what I'm leaning for:

```
      if (!fields && this === origin && _.size(this.__getAllModels(true)) == 1) {
        fields = _.keys(destination.attributes);
      }
```

which will only limit the pushed values if a model is traking all the properties and it isn't tracking another model - as is often the case when the form model is attempting to create a new object model. When using multiple models and tracking all the properties of one of them, then only push values currently existing in that model.

mandragorn:
what if you want to remove a field in the object model by unsetting it in your form model?  Do you need to explicitly set it to `null` or `undefined`?

kentmw:
When we're limiting the fields based on the existing properties in the object model, it will find the field in the object model you want to remove, and if the form model field is null, empty, undefined, or non-existent, it will set the object model's field as that.

If we aren't limiting as with the proposed setup when tracking all properties on a single model, then:
```
destination.set(this.__cloneVal(origin.attributes));
```
[...]

PR #211, opened 2016-02-27
#209, #70, #210 - FormModel refactoring

kentmw:


kentmw:
@mandragorn, the documenation (both inline and otherwise) hasn't been updated but the test cases all were updated and pass.

kentmw:
Alright, documentation up to date as well

modules/FormModel.js
```diff
@@ -47,56 +47,58 @@
...
-     *   push: {Function} a callback that will be invoked when pushing data to the Object model. It will take a single argument,
-     *     an array of all the models defined in the the model configuration array: _.pluck(computedConfig.models, 'model')
+     * mapping: {
+     *   modelName: 'foo bar baz' // track a model by providing an alias for a name and a space seperated list of fields to track as a String
+     *   modelName2: true          // to track all fields
+     *   ...                      // can have many model mappings
+     *   computedName: {
+     *     modelName: 'taz raz',  // mappings for models that will be used for this computed mapping.
+     *     ...                    // can have many model mappings for a computed
+     *     ,    // a callback that will be invoked when pulling data from the Object model
```
mandragorn: give method signature for push/pull methods

modules/FormModel.js
```diff
@@ -47,56 +47,58 @@
...
-      this.__computed = [];
       this.__cache = {};
       this.__currentUpdateEvents = [];
-      this.__modelConfigs = [];
-      this.__initMappings(options);
+      this.__currentMappings = {};
+      this.__currentObjectModels = {};
 
       // override + extend the validation and labels hashes
       this.validation = _.extend({}, this.validation || {}, options.validation || {});
```
mandragorn: do you need the || {}?  Doesn't _.extend already take care of null/undefined, etc?

modules/FormModel.js
```diff
@@ -113,86 +115,240 @@
...
+     * @return all the current mapping configs
+     */
+    getMappings: function() {
+      return this.__currentMappings;
+    },
+
+    /**
+     * Define or redefine how the form model pull/pushes or otherwise tracks properties between an object model(s).
+     * Examples:
+     * this.setMapping('modelAlias', true, optional model instance);
```
mandragorn: I forget what we said in the issue, did we want to support a shortened 'no-alias' setMapping()? Probably not, but that's the only thing that I could see that would be simpler for the common case.

Maybe something better, if you don't have a mapping for a model we push/pull everything?  So the simplest config for setting all is model.setTrackedModel(modelInstance) (i.e. no setMapping(s)())?  Or would that be confusing and unintuitive?

modules/FormModel.js
```diff
@@ -113,86 +115,240 @@
...
+     */
+    setMapping: function(alias, mapping, models, copy) {
+      var computed, fields,
+        config = {};
+      if (_.isString(mapping)) {
+        fields = mapping.split(' ');
+      } else if (mapping === true) {
+        fields = undefined;
+      } else if (_.isObject(mapping)) {
+        mapping = _.clone(mapping);
```
mandragorn: is _.clone() a deep clone (does it matter)?

modules/FormModel.js
```diff
@@ -113,86 +115,240 @@
...
+     * @method unsetTrackedModel
+     * @param aliasOrModel {String or Backbone Model instance} If a string is given, it will unset the model using that alias. If a model instance
+     *   is given, it will unbind whatever alias is currently bound to it.
      */
-    isUpdating: function() {
-      return this.__currentUpdateEvents.length > 0;
+    unsetTrackedModel: function(aliasOrModel) {
+      var alias = this.__findAlias(aliasOrModel);
+      if (alias) {
+        var model = this.__currentObjectModels[alias];
```
mandragorn: hoist?

modules/FormModel.js
```diff
@@ -255,38 +411,55 @@
...
      */
-    push: function() {
-      _.each(this.__modelConfigs, function(modelConfig) {
-        this.__copyFields(modelConfig.fields, modelConfig.model, this);
-      }, this);
-      _.each(this.__computed, function(computedConfig) {
-        // If a push callback is defined, fire it.
-        if (computedConfig.push) {
-          computedConfig.push.apply(this, [_.pluck(computedConfig.models, 'model')]);
+    isTrackingObjectModel: function() {
```
mandragorn: isTrackingAnyObjectModel?

modules/FormModel.js
```diff
@@ -255,38 +411,55 @@
...
+      return _.size(this.__currentObjectModels) > 0;
+    },
+
+    /**
+     * @method isUpdating
+     * @return true if any updates to an object model will immediately copy new values into this form model.
+     */
+    isUpdating: function() {
```
mandragorn: less sure on this one: isUpdatingAnyObjectModel()

modules/FormModel.js
```diff
@@ -255,38 +411,55 @@
...
+      return this.__currentUpdateEvents.length > 0;
+    },
+
+    /**
+     * Will add listeners that will automatically pull new updates from this form's object models.
+     * @param [pullFirst=false] {Boolean} if true, the form model will pull most recent values then start listening
+     * @method startUpdating
+     */
+    startUpdating: function(pullFirst) {
+      if (this.isTrackingObjectModel() && !this.isUpdating()) {
```
mandragorn: ahh, isUpdating() is meant to go with startUpdating().  Less sure we need the name change for isUpdating()

modules/FormModel.js
```diff
@@ -322,40 +495,111 @@
...
+     * @method __listenToModelField
+     * @private
      */
-    listenToModelField: function(model, field) {
-      var eventName = 'change:' + field;
-      this.listenTo(model, eventName, _.bind(this.__updateFormField,
-          {formModel: this, field: field}));
+    __listenToModelField: function(model, field) {
+      var eventName = field ? ('change:' + field) : 'change';
+      var callback = field ? _.bind(this.__updateFormField, {formModel: this, field: field}) : this.__updateFormModel;
```
mandragorn: there is a lot going on in this line, can it be split up for readability?

modules/FormModel.js
```diff
@@ -322,40 +495,111 @@
...
+     * @param [field] {String} the field name that it will start listening to. If no field is given, it will listen to the general 'change' event.
+     * @method __listenToModelField
+     * @private
      */
-    listenToModelField: function(model, field) {
-      var eventName = 'change:' + field;
-      this.listenTo(model, eventName, _.bind(this.__updateFormField,
-          {formModel: this, field: field}));
+    __listenToModelField: function(model, field) {
+      var eventName = field ? ('change:' + field) : 'change';
```
mandragorn: Just writing it to see how it looks... hoisting is annoying:
```
field = field ? ':' + field : '';
var eventName = 'change' + field;
```

what you have already is probably clearer.

modules/FormModel.js
```diff
@@ -322,40 +495,111 @@
...
+        var model = this.getTrackedModel(modelAlias);
+        if (model) {
+          models[modelAlias] = model;
+        } else {
+          hasAllModels = false;
+        }
+      }, this);
+      return hasAllModels ? models : undefined;
+    },
+
```
mandragorn: is there value in having a general debugging method that will return the state of the formModel?  Existing mappings and defined or undefined models in an easily digestible form when debugging?  Its getting complex enough that having a simple health check could be useful.

modules/FormModel.js
```diff
@@ -676,74 +961,109 @@
...
       if (model) {
         this.formModel.__updateCache(model);
       }
-      (function(formModel, pullCallback, modelConfigs) {
-        _.each(modelConfigs, function(modelConfig) {
-          if (modelConfig.fields) {
-            _.each(modelConfig.fields, function(field) {
-              args.push(formModel.__cloneVal(modelConfig.model.get(field)));
-            });
+      (function(formModel, alias) {
```
mandragorn: I don't fully understand why the function definition and call is necessary.

modules/FormModel.js
```diff
@@ -322,40 +495,111 @@
...
+     * @method __listenToComputedValuesDependency
+     * @private
      */
-    listenToComputedValuesDependency: function(computedConfig, model, field) {
-      var eventName = 'change:' + field;
-      this.listenTo(model, 'change:' + field, _.bind(this.__invokeComputedPull,
-          {formModel: this, models: computedConfig.models, pull: computedConfig.pull}));
+    __listenToComputedValuesDependency: function(model, field, computedAlias) {
+      var eventName = field ? ('change:' + field) : 'change';
+      this.listenTo(model, eventName, _.bind(this.__invokeComputedPull,
```
mandragorn: why _.bind(this.__invokeComputedPull, {...}) instead of _.bind(this.invokeComputedPull, this, eventName)?

modules/FormModel.js
mandragorn: I think I'm thinking of curry, not bind.

modules/FormModel.js
mandragorn: I think I'm thinking of curry, not bind.

modules/FormModel.js
```diff
@@ -322,40 +495,111 @@
...
+     */
+    __getModelAliases: function(computedAliasOrConfig) {
+      var config,
+        modelAliases = [];
+      if (_.isString(computedAliasOrConfig)) {
+        config = this.getMapping(computedAliasOrConfig);
+      } else {
+        config = computedAliasOrConfig;
+      }
+      _.each(config.mapping, function(mapping, key) {
```
mandragorn: modelAliases = _.map(config.mapping, function(mapping, key) { ... });

writing it out with chaining to see how it looks:
```
modelAliases = _.chain(config.mapping)
    .map(function(mapping, key) {
      return key;
    })
    .filter(function(key) {
      return key != 'pull' && key != 'push';
    })
    .value();
```

nope, not better to do chaining.

modules/FormModel.js
```diff
@@ -438,12 +682,54 @@
...
+     * @method __push
+     * @param alias {String} the name of the mapping that will be used during the push
+     * @private
+     */
+    __push: function(alias) {
+      var config = this.getMapping(alias);
+      if (config.computed && config.mapping.push) {
+        var models = this.__getComputedModels(alias);
+        if (models) {
+          config.mapping.push.call(this, models);
```
mandragorn: what if push isn't defined for the mapping? Is a default one created?

modules/FormModel.js
```diff
@@ -676,74 +961,109 @@
...
-            _.each(modelConfig.fields, function(field) {
-              args.push(formModel.__cloneVal(modelConfig.model.get(field)));
-            });
+      (function(formModel, alias) {
+        var hasAllModels = true,
+          config = formModel.getMapping(alias),
+          modelAliases = formModel.__getModelAliases(alias),
+          models = {};
+        if (!config.mapping.pull) {
+          console.log('Not pulling the computed: ' + alias + ', because no pull method was defined for this computed.');
```
mandragorn: is console.log safe to use in IE?

modules/FormModel.js
```diff
@@ -538,23 +852,33 @@
...
     __setupListeners: function() {
-      _.each(this.__modelConfigs, function(modelConfig) {
-        if (modelConfig.fields) {
-          _.each(modelConfig.fields, function(field) {
-            this.listenToModelField(modelConfig.model, field);
-          }, this);
+      var formModel = this;
+      _.each(formModel.getMappings(), function(config, alias) {
+        if (config.computed) {
+          var modelConfigs = formModel.__getComputedModelConfigs(alias);
```
mandragorn: hoist?

modules/FormModel.js
```diff
@@ -538,23 +852,33 @@
...
+          _.each(modelConfigs, function(modelConfig) {
+            var model = modelConfig.model;
+            if (modelConfig.fields) {
+              _.each(modelConfig.fields, function(field) {
+                formModel.__listenToComputedValuesDependency(model, field, alias);
+              });
+            } else {
+              formModel.__listenToComputedValuesDependency(model, undefined, alias);
```
mandragorn: ugly params.  Switch to single object with named params?

modules/FormModel.js
```diff
@@ -581,69 +907,28 @@
...
-          model: mapping.model,
-          fields: mapping.fields
-        });
-      } else if (mapping && mapping.models) {
-        modelConfigs = mapping.models.slice();
+    __initMappings: function(options) {
+      var mapping,
+        defaultMapping = _.result(this, 'mapping'),
+        optionsMapping = options.mapping;
+      mapping = options.mapping || defaultMapping;
```
mandragorn: not worrying about merging them?  options.mapping key/values override defaultMapping key/values?  Instead of using one or the other.

modules/FormModel.js
```diff
@@ -676,74 +961,109 @@
...
+          var fields = config.mapping[modelAlias],
+            model = formModel.getTrackedModel(modelAlias),
+            modelCopy = {};
+          if (!model) {
+            hasAllModels = false;
           } else {
-            args.push(formModel.__cloneVal(modelConfig.model.attributes));
+            if (fields) {
+              _.each(fields, function(field) {
+                modelCopy[field] = formModel.__cloneVal(model.get(field));
```
mandragorn: is __cloneVal() a deep clone?

mandragorn:
got through the code, not the tests.

modules/FormModel.js
kentmw: done - removed

modules/FormModel.js
kentmw: it's a shallow copy and i don't believe it matters. deep cloning things with functions is more trouble then i expect it to be for others.

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: nah, i don't think this one makes that much sense. I think the added verbosity is not as useful

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: split it up

modules/FormModel.js
kentmw: are this.getMappings() and this.getModels() not good enough for debugging purposes?

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: because backbone already provides the first (and second i think) argument to event listener callbacks on model changes

modules/FormModel.js
kentmw: this is the callback for an event listener on changes of a model. Backbone will invoke this method for us typically with the context of the form model and the argument being the model changed. I need more information, specifically the alias for the computed mapping. Thus we change or add to the context of the callback to include this information

modules/FormModel.js
kentmw: done - not changing

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: i imagine that the version of jquery we are using as a hard dependency requires a version of IE that has console.log...

modules/FormModel.js
kentmw: Yes - minus functions which should not be on models anyway

modules/FormModel.js
kentmw: I thought about it, then decided it was more pain to do right now than necessary - specifically how you remove a mapping on a parent. We can add it later if needed? In the test formModelInitialization I show how you can use the parent's prototype to extend it and override values. I don't know. I'm on the fence and I don't have a lot of brain power or motiviation right now to think about making it better. If you have suggestings or feelings, let me know

modules/FormModel.js
kentmw: then __listenToModelField wouldn't parallel this function anymore... hmm... Does formModel.__listenToComputedValuesDependency(model, '', alias); look better? it's still a falsey value.

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: done

modules/FormModel.js
kentmw: Everytime I want to simplify the base cases I find that the fundamentals / structure of the mapping vs tracked models becomes cloudy or stretched. No-alias mappings don't make sense after they've been set. How would you change or remove this mapping but using the alias - you'd have to store the auto-generated value anyway. And adding a model.setTrackedModel(instance) that generates a mapping confuses the separation of mapping and tracked models. Currently to do what you are proposing model.setTrackedModel(instance) does would be:
model.setMapping('myModel', true, instance). In your opinion, is this bad enough to add a convenience method that blurs the separation of mapping and tracked models?

modules/FormModel.js
mandragorn: The console object is only available in IE9 when the developer tools are open.
jQuery 2.x supports IE 9+

So in IE 9 using console will break things.

kentmw: Wow, learned something new. Thanks for sharing that. I can check for
console and console.log

modules/FormModel.js
mandragorn: ok, this looks like a good balance for now.

modules/FormModel.js
mandragorn: I was thinking something that makes it more obvious what models aren't defined.  But I realize now I'm thinking about super complicated cases where there are more than a handful of models, which is rarely (if ever) the case.

modules/FormModel.js
mandragorn: the change to just _.map isn't clearer (my comment was that the full chaining thing wasn't clearer, you can still do it with just a map)?  Or is there an issue where _.map includes null/undefined items?

modules/FormModel.js
mandragorn: ok

mandragorn:
+1 - I'd like Ben Pepper or someone else from the front-end team to review this to make sure it is in line with their needs.

modules/FormModel.js
kentmw: I went with:
```     
 return _.filter(_.keys(config.mapping), function(key) {
   return key != 'pull' && key != 'push';
 });
```

modules/FormModel.js
kentmw: done

kentmw:
Alright @pepperbc @mandragorn I'm going to merge this tomorrow afternoon. Please add notes if you want changes before then.

kentmw:
Please provide comments about future updates if you have any

Issue #128, opened 2015-09-14
Torso Yuidocs


mandragorn:
As a developer I would like easy access to the full code documentation of Torso.

Acceptance Criteria:
* Demonstrate that you can view the documentation of a specific version of Torso without downloading and/or building the documentation locally.

kentmw:
In the .gitignore we are ignoring the docs/js

Issue #78, opened 2015-06-22
Change collectionLoadingMixin to just loadingMixin


kentmw:
I'm pretty sure it works for models as well. Should maybe change that.

PR #216, opened 2016-03-08
#128, #78, Updated naming of mixins, moved them to their own director…

kentmw:
…y, added js docs as a tracked directory, fixed lint errors, and updated the demo.


mandragorn:
high level +2

Issue #217, opened 2016-03-08
Upgrade Backbone-Nested to 2.0.4


kentmw:
There's an issue with delayedTriggers being a singleton.

Issue #202, opened 2016-02-20
Cache collections use JSON.stringify on ids when fetchingByIds


kentmw:
This absolutely destroys being able to send ids as query parameters when using a GET request

PR #218, opened 2016-03-08
#217, #202 - Upgrade backbone-nested to 2.0.4, not stringifying idsTo…

kentmw:
…Fetch on cache fetch, and removed unneeded test console logs


mandragorn:
looks good to me.

Issue #219, opened 2016-03-10
IE not happy with document.activeElement call in hotswapKeepCaret when in an iFrame


maurag13:
We currently have hotswapKeepCaret starting with this:

var currentCaret,
          activeElement = document.activeElement;

which works fine from iFrames in Chrome & Firefox, but not so much in IE (https://www.tjvantoll.com/2013/08/30/bugs-with-document-activeelement-in-internet-explorer/).  Needs a try/catch wrapped around the activeElement call.

mandragorn:
fixed in #220

PR #220, opened 2016-03-10
getting hotswapKeepCaret to work with IE & iFrames

maurag13:
document.activeElement doesn't play well with IE & iFrames, so we need to wrap it in a try/catch.  See issue #219

### DIFF
--- a/modules/templateRenderer.js
+++ b/modules/templateRenderer.js
@@ -182,6 +182,9 @@
       } catch (error) {
         activeElement = null;
       }
+      if (activeElement && this.supportsSelection(activeElement)) {
+        currentCaret = this.getCaretPosition(activeElement);
+      }
       this.hotswap(currentNode, newNode, ignoreElements);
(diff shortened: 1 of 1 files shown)


modules/templateRenderer.js
```diff
@@ -176,9 +176,11 @@
 
     hotswapKeepCaret: function(currentNode, newNode, ignoreElements) {
       var currentCaret,
-          activeElement = document.activeElement;
-      if (activeElement && this.supportsSelection(activeElement)) {
-        currentCaret = this.getCaretPosition(activeElement);
+          activeElement;
+      try {
+        activeElement = document.activeElement;
```
arikwex: but... what about keeping the caret position? :(

maurag13: Gah - deleted more lines than I meant to!  Thank you for catching; I will fix immediately.

mandragorn:
+2

PR #225, opened 2016-03-15
#128:  Using yuidoc cli

mandragorn:
converting yuidoc to use yuidoc cli instead of the gulp plugin.


kentmw:
Can't really tell, but +1

package.json
```diff
@@ -81,5 +82,14 @@
       "name": "Josh Young",
       "email": "jyoung@vecna.com"
     }
-  ]
+  ],
+  "yuidoc": {
+    "name": "backbone-torso",
+    "description": "backbone-torso",
```
mandragorn: what should I put in here?

Issue #230, opened 2016-03-18
collection.js tests are broken


kentmw:
The mockjax file was not updated to reflect the changes in the the fetchByIds payload for GET requests.

Issue #228, opened 2016-03-17
Remove ListView's noop prepare method override


kentmw:
Why remove the view state from the context?

mandragorn:
I don't know, it was very confusing when I came across it.

kentmw:
It was added when the prepare method only added the model and a list view most likely doesn't have a model

Issue #227, opened 2016-03-16
FormView should have a way of resetting FormModel and update listeners.


kentmw:
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/FormView.js#L91-L94

Setting up listeners on the form model has to be performed manually as of now when you update the form model instance. There should be a method to help reset the form model

PR #231, opened 2016-03-18
#230, #228, #227 - list view's override of prepare method removed, co…

kentmw:
…llection.js tests fixed, and form view has a resetModel method now


kentmw:
@diokey and @mandragorn this PR will hopefully hit some of the issues we talked about.

Issue #240, opened 2016-04-11
Shared code / behaviors enhancement


kentmw:
How do you share behaviors between two Torso views? For example:

```
var EditInsurancePerspective = TorsoView.extend({
  template: '<div inject="modal" />',

  events: {
    'click .back-link': '_openConfirmationModal'
  },

  initialize: function() {
    this._confirmationModal = new ConfirmationModal({
      text: 'Confirm exit Insurance edit'
    });
    this.set('confirmModalIsOpen', false);
    this.listenTo(this.viewState, 'change:confirmModalIsOpen', this.render);
  },

  attachTrackedViews: function() {
    if (this.get('confirmModalIsOpen')) {
      this.attachView('modal', this._confirmationModal);
    }
  },

  _openConfirmationModal: function() {
    this.set('confirmModalIsOpen', true);
  }
});

var EditDemographicsPerspective = TorsoView.extend({
  template: '<div inject="modal" />',

  events: {
    'click .back-link': '_openConfirmationModal'
  },

  initialize: function() {
    this._confirmationModal = new ConfirmationModal({
      text: 'Confirm exit Demographics edit'
    });
    this.set('confirmModalIsOpen', false);
    this.listenTo(this.viewState, 'change:confirmModalIsOpen', this.render);
  },

  attachTrackedViews: function() {
    if (this.get('confirmModalIsOpen')) {
      this.attachView('modal', this._confirmationModal);
    }
  },

  _openConfirmationModal: function() {
    this.set('confirmModalIsOpen', true);
  }
});
```

We want to share multiple behaviors across many views that aren't part of the same inheritance hierarchy.

One implementation:

```
var Behaviors = {};

var eventMap = {
  attached: '_attach',
  detached:  '_detach',
  activated: '_activate',
  deactivated: '_deactivate',
  disposed: '_dispose',
  attachedTrackedViews: '_attachTrackedViews',
  'render:begin': '_prerender', 
  'render:complete': '_postrender',
  updatedDOM: '_updateDOM',
  initialize:begin:  '_preinitialize',
  'initialize:complete': '_postinitialize'
};

var Torso.Behavior = Torso.Cell.extend({
  constructor: function(args) {
    this.view = args.view;
    this.config =  _.defaults(args.configOverrides, this.defaults);
    this.setupCallbacks();
    this.bindViewEvents();
  },

  action: _.noop,

  setupCallbacks: function() {
    var actionCallback = _.bind(this.action, this);
[...]

PR #246, opened 2016-04-15
Ref #240: First draft of Behavior implementation

Earthstar:
MVP of Behavior implementation

Behavior lifecycle hooks fire at the appropriate times
Behaviors can define events in the event hash
Views can define Behaviors


test/karma/behaviorSpec.js
```diff
@@ -0,0 +1,522 @@
...
+
+    it('_dispose before view._dispose', function() {
+      var methodsCalled = [];
+      var ViewTrackingDispose = generateViewAndBehaviorTracking(methodsCalled, '_dispose');
+      var viewTrackingDispose = new ViewTrackingDispose();
+      viewTrackingDispose.dispose();
+
+      expect(methodsCalled[0]).toBe('behavior:_dispose');
+      expect(methodsCalled[1]).toBe('view:_dispose');
+
```
Earthstar: I could write the tests using auto-generated classes and custom matchers. The advantage is that you'd have less test boilerplate and better error messages, but it's harder to debug. Thoughts?

test/karma/behaviorSpec.js
```diff
@@ -0,0 +1,522 @@
...
+        });
+
+        var viewWithBehaviorAndEvent = new ViewWithBehaviorAndEvent();
+
+        expect(savedContext).toBe(viewWithBehaviorAndEvent);
+      });
+    });
+
+    it('if duplicate events are defined in view and behavior event hash, only view handler runs', function() {
+      var ViewWithBehaviorAndEvent = TorsoView.extend({
```
Earthstar: Will we ever want both handlers to run? Do we want to warn people if they're overriding something?

modules/Behavior.js
```diff
@@ -0,0 +1,66 @@
...
+
+  _bindLifecycleMethods: function() {
+    _.each(eventMap, function(callback, event) {
+      this.listenTo(this.view, event, this[callback]);
+    }, this);
+  },
+
+  _bindEventCallbacks: function() {
+    var behaviorEvents = _.result(this, 'events');
+    var viewEvents = this.view && this.view.events;
```
Earthstar: Should Behavior throw an exception if it's being constructed without a view?

modules/View.js
```diff
@@ -78,6 +79,28 @@
...
+    _initializeBehaviors: function() {
+      var self = this;
+      if (!_.isEmpty(this.behaviors)) {
+        self.__behaviorInstances = {};
+        _.each(this.behaviors, function(behaviorDefinition, alias) {
+          var BehaviorClass = behaviorDefinition.behavior;
+          var behaviorOptions = _.pick(behaviorDefinition, function(value, key) {
+            return key !== 'behavior';
+          });
+          // Is it okay to override potential 'view' parameters?
```
kentmw: yes, remove this comment.

modules/View.js
```diff
@@ -368,6 +394,7 @@
      * @method activate
      */
     activate: function() {
+      this.trigger('before-activate');
```
kentmw: i wonder if it's better to trigger it right before this._activate...

really on the fence.

modules/View.js
```diff
@@ -78,6 +79,28 @@
...
+      if (!_.isEmpty(this.behaviors)) {
+        self.__behaviorInstances = {};
+        _.each(this.behaviors, function(behaviorDefinition, alias) {
+          var BehaviorClass = behaviorDefinition.behavior;
+          var behaviorOptions = _.pick(behaviorDefinition, function(value, key) {
+            return key !== 'behavior';
+          });
+          // Is it okay to override potential 'view' parameters?
+          behaviorOptions.view = self;
+          self.__behaviorInstances[alias] = new BehaviorClass(behaviorOptions);
```
Earthstar: Should throw a more specific error message here if the Behavior definition is wrong.

modules/Behavior.js
```diff
@@ -0,0 +1,66 @@
...
+module.exports = TorsoCell.extend({
+  constructor: function(options) {
+    options = options || {};
+    this.view = options.view;
+    this._bindLifecycleMethods();
+    this._bindEventCallbacks();
+    TorsoCell.apply(this, arguments);
+  },
+
+  _bindLifecycleMethods: function() {
```
kentmw: I think these private methods should be `__` prefixed instead of `_` as they are Torso private methods.

test/karma/behaviorSpec.js
```diff
@@ -0,0 +1,522 @@
+var _ = require('underscore');
+var TorsoBehavior = require('./../../modules/Behavior');
+var TorsoView = require('./../../modules/View');
+
+var spyOnBackbone = require('./helpers/spyOnBackbone');
```
Earthstar: You can't use spyOnBackbone as currently implemented to track things that happen during initialize, because it wraps the "initialize" method.

test/karma/behaviorSpec.js
```diff
@@ -0,0 +1,522 @@
...
+    var firstViewWithBehavior = new ViewWithBehavior();
+    var secondViewWithBehavior = new ViewWithBehavior();
+
+    var firstBehavior = firstViewWithBehavior.getBehavior('torsoBehavior');
+    var secondBehavior = secondViewWithBehavior.getBehavior('torsoBehavior');
+
+    expect(firstBehavior).not.toBe(secondBehavior);
+  });
+
+  it('A view can have multiple behaviors distinguished by alias', function() {
```
Earthstar: May wish to add specifications around when the lifecycle methods of multiple behaviors run.

test/karma/behaviorSpec.js
kentmw: oh man, this isn't great. I was super ready to say, yes! the view wins. But the event keys here are about dom elements and their related dom events. If a behavior is listening to a click on a button, the view should definitely be able to listen to a click on that button too.

modules/Behavior.js
kentmw: yes.

modules/View.js
Earthstar: If we really want to trigger events right before and after lifecycle hooks, would it be a good idea to wrap the hooks with a function that automatically generates/triggers the events, similar to Java AOP? Advantage: less duplication, less chance of forgetting to add an event. Disadvantage: harder to understand.

test/karma/behaviorSpec.js
Earthstar: I'm leaning towards having both handlers run as well. The implementation right now is awkward because it's dependent on the event hash keys being the same. So, you could get two handlers to run on the same element if they used different selectors. If we wanted to allow both to run, we'd have to re-think how to implement registering the behavior events. Is this required for MVP?

test/karma/behaviorSpec.js
kentmw: I think that the order that you define the behviors is the order they should run in if they bind to the same lifecycle method

modules/View.js
```diff
@@ -78,6 +79,28 @@
       return this.viewState.set.apply(this.viewState, arguments);
     },
 
+    _initializeBehaviors: function() {
```
mandragorn: docs?

modules/Behavior.js
```diff
@@ -0,0 +1,66 @@
...
+      if (!behaviorEvents) {
+        return;
+      } else {
+        viewEvents = {};
+      }
+    }
+
+    var boundBehaviorEvents = this._bindEventCallbacksToBehavior(behaviorEvents);
+
+    if (_.isFunction(viewEvents)) {
```
mandragorn: does this block duplicate what backbone does to bind view events?

Earthstar: A bit. In View.bindEvents, the View will attempt to bind the behavior event callbacks to the view's context, although when you bind a method multiple times, the first bind wins. The logic in _bindEventCallbacksToBehavior to find the method is also duplicated. I don't see a way of sharing the logic without modifying Backbone.

kentmw: I don't see it as a duplication of backbone code. It's just using Backbone's work downstream and plugging into (extending) its events hash upstream.

test/karma/behaviorSpec.js
Earthstar: This bears out in testing.

test/karma/behaviorSpec.js
Earthstar: Decided not to auto-generate classes

test/karma/behaviorSpec.js
```diff
@@ -189,26 +158,57 @@ describe('A Torso Behavior', function() {
...
-    var behaviorWithInitialize = viewWithBehaviorWithInitialize.getBehavior('behaviorWithInitialize');
-    expect(behaviorWithInitialize.initializeArguments).toBeDefined();
-    expect(_.isMatch(behaviorWithInitialize.initializeArguments[0])).toBe(true);
+    var behaviorRecordingInitializeArguments = viewWithBehaviorWithInitialize.getBehavior('behaviorRecordingInitializeArguments');
+    expect(behaviorRecordingInitializeArguments.initializeArguments).toBeDefined();
+    var recordedBehaviorOptions = behaviorRecordingInitializeArguments.initializeArguments[0];
+    expect(_.isMatch(recordedBehaviorOptions, {propertyKey: 'propertyValue'})).toBe(true);
+  });
+
+  it('view options are passed to behavior initialize', function() {
```
Earthstar: While porting over an existing mixin to a Behavior, I realized Behaviors need access to a View's initialization options. Right now, I'm merging the options object passed into a View's initialization with the Behavior's options, but I don't like this because I think overriding properties is unintuitive. I'd prefer to either save the view's initialization options on the view, or pass in the options as a separate object. Thoughts?

kentmw: I'd either go with merging them together in a separate object `_.extend({}, args, behaviorArgs)` or invoke the initialization of the behavior with 2 arguments (viewArgs and behaviorsArgs)

modules/Behavior.js
```diff
@@ -0,0 +1,146 @@
...
+     * @method __namespaceEvents
+     * @param eventHash {Object} to namespace
+     * @return {Object} with event namespaced with '.behavior' and the cid of the behavior
+     * @private
+     */
+    __namespaceEvents: function(eventHash) {
+      // coped from Backbone
+      var delegateEventSplitter = /^(\S+)\s*(.*)$/;
+      var namespacedEvents = {};
+      var behaviorId = this.cid;
```
kentmw: is this better than grabbing the view's alias for this behavior?

test/karma/behaviorSpec.js
Earthstar: I prefer having two arguments because I don't want to bother with overriding keys.

modules/Behavior.js
Earthstar: I went with a cid because I don't think the behavior should know or care what the view decides to name the behavior. I could add the view's name for the behavior as an additional namespace, but I'm not sure what value that provides. Example?

kentmw: Just readability when someone inspects the bound events

kentmw:
+2 everything we spoke about can be updated afterwards as we see it necessary.

Issue #249, opened 2016-04-26
release-it fails with missing unprefixed version


kentmw:
It dies when I try to "release-it":

```
WARNING Probably the current version in package.json is not a known tag in the repository.
ERROR Could not create changelog from latest tag (0.6.5) to HEAD.
```

It's because the last release wasn't tagged ever as 0.6.5. @mandragorn, thoughts?

mandragorn:
yea, I had to modify things to add the v to the tag since you were doing that before.  If we drop the v from the tag name the issues should go away.  The main reason I wanted to standardize by using release-it was to avoid the inconsistent tag names, etc.

PR #250, opened 2016-04-26
#249: Updating identification of previous tag for release notes.

mandragorn:


mandragorn:
will merge and test releasing.  Can rollback or revert if it doesn't work.

mandragorn:
fixes #249

Issue #241, opened 2016-04-11
Look into peerDependencies as a potential for Backbone and jQuery


kentmw:
Just saw peerDependencies were a thing in npm. I need to look into them to see if they're something we could use for Backbone and jQuery which I think are peer dependencies.

Earthstar:
Desired behavior: Torso does not bring in its own jQuery, etc. dependency when installed in another project. 

Based on some research, I think simply adding jQuery as a peer dependency will work. npm < 3 will automatically install jQuery. If we want to support npm > 3, we'd probably need to additionally specify jquery as a devDependency so Torso can continue to run tests.

PR #259, opened 2016-05-13
Ref #241: Change libraries to peer dependencies

Earthstar:
I think this will work, but I'm not sure how to test it properly without releasing a package. I've duplicated peerDependencies in devDependencies for future-compatibility should we switch to npm > 3.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,14 +17,17 @@
     "clean": "gulp clean"
   },
   "main": "modules/torso.js",
+  "peerDependencies": {
+    "jquery": "2.1.x",
+    "backbone": "1.2.x",
+    "underscore": "1.8.x"
+  },
   "dependencies": {
-    "jquery": "2.1.3",
-    "underscore": "1.8.2",
-    "backbone": "1.2.3",
     "backbone-nested": "2.0.4",
     "backbone.stickit": "0.9.2"
   },
   "devDependencies": {
+    "backbone": "1.2.3",
     "browserify": "^12.0.2",
     "del": "^1.1.1",
     "gulp": "^3.8.8",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -17,14 +17,18 @@
...
+  },
   "dependencies": {
-    "jquery": "2.1.3",
-    "underscore": "1.8.2",
-    "backbone": "1.2.3",
     "backbone-nested": "2.0.4",
     "backbone.stickit": "0.9.2"
   },
   "devDependencies": {
+    "backbone": "1.2.3",
```
mandragorn: should these dependencies also follow *.*.x ?  If we expect it to work with peerDependencies, then we probably want to know as soon as possible when the latest version of a dependency breaks things.

package.json
```diff
@@ -17,14 +17,18 @@
...
+    "jquery": "2.1.x",
+    "backbone": "1.2.x",
+    "underscore": "1.8.x",
+    "handlebars": "3.0.x"
+  },
   "dependencies": {
-    "jquery": "2.1.3",
-    "underscore": "1.8.2",
-    "backbone": "1.2.3",
     "backbone-nested": "2.0.4",
```
Earthstar: Fortunately, backbone-nested and backbone.stickit don't bring in jquery, underscore, or backbone when installed. Because they use Bower.

mandragorn:
:+1:

kentmw:
Is this ready to merge?

Earthstar:
Yes, if you think all three should be peer dependencies. I think figuring
out the versions of jquery, etc Torso supports is a whole issue in and of
itself.

Deena

kentmw:
follow up: remove handlebars from a peer dependency. Torso doesn't depend on handlebars

PR #261, opened 2016-06-16
fixes #260 - FormModel set simple mapping with the copy flag

### LINKED ISSUE #260: FormModel.setMapping() doesn't copy when the copy parameter is set.
```
model.setMapping('test', true, testModel, true);
```
does not copy when the mapping is set.

This line: https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/FormModel.js#L194
needs to move before setting the tracked model so that the mappings exist when it tries to copy.

mandragorn:


kentmw:
Don't know what it does but looks good

mandragorn:
The add tracked models has a property to perform a copy which it does based on the mappings.  If the mappings aren't set then when it does the copy it copies nothing.  This just sets the mappings prior to tracking (and copying) the models.

PR #264, opened 2016-06-23
fixes #263 - updating pull to keep track of requested ids that are in-flight

### LINKED ISSUE #263: Pull requests do not take into account stuff that is being fetched, but hasn't returned.
If you have a private collection and do pull requests for the same ids in quick succession then each request will perform duplicate fetches to the server.

This is because the models are not yet in the cache so pull doesn't know to exclude those ids.

Possible fix: keep track of requestIds being fetched and don't pull those.
Here is where pull should be taking into account ids of models in the process of being fetched: https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/mixins/cacheMixin.js#L140

Here is where we need to track the requested ids, and then forget about them when the request completes (success or failure): https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/mixins/cacheMixin.js#L347-L389

mandragorn:


mandragorn:
@catbieber - more CS tasking for ya...

modules/mixins/cacheMixin.js
```diff
@@ -138,6 +140,7 @@
             options = options || {};
             //find ids that we don't have in cache
             options.idsToFetch = _.difference(this.getTrackedIds(), _.pluck(parentInstance.models, 'id'));
+            options.idsToFetch = _.difference(options.idsToFetch, _.uniq(_.flatten(_.values(parentInstance.idsFetching))));
```
maurag13: Isn't this line just overwriting what's in idsToFetch, so you should delete line 142?

torso-bundle.js
```diff
@@ -681,6 +683,7 @@
             options = options || {};
             //find ids that we don't have in cache
             options.idsToFetch = _.difference(this.getTrackedIds(), _.pluck(parentInstance.models, 'id'));
+            options.idsToFetch = _.difference(options.idsToFetch, _.uniq(_.flatten(_.values(parentInstance.idsFetching))));
```
maurag13: Same question about whether you're trying to overwrite or append to options.idsToFetch.

maurag13:
+1

modules/mixins/cacheMixin.js
```diff
@@ -351,10 +355,14 @@
...
               data: {ids: requestedIds.join(',')}
             };
-          if (options.fetchContentType || (ajaxOpts.type && ajaxOpts.type.toUpperCase() != 'GET')) {
-            ajaxOpts.contentType = options.fetchContentType || 'application/json; charset=utf-8';
+          if (contentType || (ajaxOpts.type && ajaxOpts.type.toUpperCase() != 'GET')) {
+            ajaxOpts.contentType = contentType || 'application/json; charset=utf-8';
             ajaxOpts.data = JSON.stringify(requestedIds);
           }
+          if (!collection.idsFetching) {
+            collection.idsFetching = {};
```
littleice: should the collection.idsFetching be an array [] instead of {}?

PR #265, opened 2016-06-24
Reverting peer dependency for backbone because it screws with existin…

mandragorn:
…g builds using torso.


mandragorn:
merging so I can test since testing w/o tagging is proving unreliable.

PR #269, opened 2016-07-15
Have the loadingMixin loading status be updated before triggering load-complete

maurag13:
We have some logic where we listen for a few load-complete events to kick off a method, and then in that method look to see if the .loading value for one of the objects is true or false.  Since the always() was after the trigger of load-complete, this meant that we could sometimes hit a race condition where we run the method before we get to the always().  I believe that having the always() logic come first will both fix this situation and not harm other existing logic.


maurag13:
@catbieber & @kentmw  consider yourself tagged  :-)

PR #274, opened 2016-08-04
Adding initialize event triggers on view so that they can be used by …

mandragorn:
…the existing code in Behavior.js


kentmw:
+2

PR #276, opened 2016-08-04
Fixing initial values for some view properties so they can be overrid…

mandragorn:
…den by behaviors.


kentmw:
huh, what's the difference?

mandragorn:
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/View.js#L686 explicitly checks for undefined.  null is not undefined.

mandragorn:
merging so I can release.  We can revert if needed.

PR #277, opened 2016-08-12
Adding prepare() to Behaviors.

mandragorn:


modules/Behavior.js
```diff
@@ -23,7 +23,7 @@
     'render:begin': 'prerender',
     'render:complete': 'postrender',
     'initialize:begin':  '_preinitialize',
-    'initialize:complete': '_postinitialize'
+    'initialize:complete': '__postinitialize'
```
kentmw: don't change this, keep it as a list of hooks users can use

kentmw: Add a special listener to initialize:complete that does what __postinitialize does.

modules/Behavior.js
```diff
@@ -49,10 +53,18 @@
     mixin: {},
 
     /**
+     * The behavior's prepare result will be combined with the view's prepare with the behavior's alias as the namespace.
+     * effectively: { <behaviorName>: behavior.prepare() } will be combined with the view's prepare result.
+     * @method prepare
+     * @return a prepare context suitable to being added to the view's prepare result.
+     */
+    prepare: _.noop,
```
kentmw: grr... i wish that prepare was part of the lifecycle events list. now it's harder to find.

modules/View.js
```diff
@@ -680,6 +680,7 @@
           });
           behaviorOptions.view = view;
           var behaviorInstance = view.__behaviorInstances[alias] = new BehaviorClass(behaviorOptions, viewOptions);
+          behaviorInstance.alias = alias;
```
mandragorn: add the alias to behaviorOptions above and set it in the constructor so its available to initialize.

modules/Behavior.js
mandragorn: update to postinitialize calling _postinitialize internally.

modules/Behavior.js
```diff
@@ -23,7 +23,7 @@
     'render:begin': 'prerender',
     'render:complete': 'postrender',
     'initialize:begin':  '_preinitialize',
```
mandragorn: update to preinitialize using same logic as postinitialize

modules/Behavior.js
```diff
@@ -68,6 +80,47 @@
...
+     * such that it returns the combination of the view's prepare result with the behavior's prepare result
+     * inside it under the behavior's alias.
+     * This method will call _postinitialize when its finished.
+     * @method __postinitialize
+     * @private
+     */
+    __postinitialize: function() {
+      var originalViewPrepareFn = _.bind(this.view.prepare, this.view);
+      var wrappedPrepareFn = _.wrap(originalViewPrepareFn, this.__viewPrepareWrapper);
+      this.view.prepare = _.bind(wrappedPrepareFn, this);
```
kentmw: do we want the bound/wrapped function to have the behavior as the context of the view? I would have thought the view

kentmw:
+2

PR #280, opened 2016-08-13
Fixing arguments to Behaviors (Cells).

mandragorn:
The first argument to cells is the attributes, not the options.  So with the way they are currently coded the behaviorOptions end up as attributes of the cell.  This could have performance impacts if the behaviors take options that are complex (since a behavior is a Cell).

Why don't behaviors extend NestedCell?


mandragorn:
@kentmw this is ready for re-review.

kentmw:
+2

modules/Behavior.js
```diff
@@ -63,12 +63,13 @@
...
      * @method constructor
      * @override
+     * @param behaviorAttributes {Object} the initial value of the behavior's attributes.
      * @param behaviorOptions {Object}
      *   @param behaviorOptions.view {Backbone.View} that Behavior is attached to
      *   @param behaviorOptions.alias {Backbone.View} the alias for the behavior in this view.
      * @param [viewOptions] {Object} options passed to View's initialize
      */
-    constructor: function(behaviorOptions, viewOptions) {
+    constructor: function(behaviorAttributes, behaviorOptions, viewOptions) {
```
mandragorn: this is not backwards compatible and I don't see a good way to make it backwards compatible w/o going back to what it was.

kentmw: yep, stupid cells/models taking attributes first.

Issue #282, opened 2016-08-23
I would like a simple syntax for interacting with my data objects (Data Utils)


mandragorn:
Should it be a behavior or built-in to the View?

If behavior, one behavior per data object or one behavior per view?

mandragorn:
![torso data utils software engineering vecna shared documents](https://cloud.githubusercontent.com/assets/807561/17877103/51d7528a-68ae-11e6-87f6-421802d16b6c.png)

mandragorn:
From discussion on 8/22:
* Use a behavior
* some api changes
** cache: cacheCollectionInstance
** isCollection: true (or type: 'collection' / 'model')
* Trigger change events on behavior and have dependencies be between behaviors
* prepare of behaviors is nice.

mandragorn:
The result's (collection or model) toJSON() is added to the view's prepare namespaced by the behavior's alias.

i.e. a behavior with the alias demographics that is model-typed will add the demographics model's .toJSON() value to the view's template context as:
```
viewTemplateContext = {
  ...
  demographics: { ... }
  ...
}
```

a behavior with the alias appointments that is collection-typed will add the appointment collection's .toJSON() value to the view's template context as:
```
viewTemplateContext = {
  ...
  appointments: [ { ... }, { ... }, ... ]
  ...
}
```


Options:
* __cache__ - the torso collection that is acting as a cache used to create the private collections.  ignored if __model__ is set.
* __model__ - a 'singleton' model that will be managed by this behavior.
* __type__ - one of TorsoDataBehavior.TYPES.MODEL or TorsoDataBehavior.TYPES.COLLECTION, defaults to TorsoDataBehavior.TYPES.COLLECTION if not set.  Ignored if __model__ is set.
* __fetchMode__ - one of TorsoDataBehavior.FETCH_MODES.PULL or TorsoDataBehavior.FETCH_MODES.FETCH.  defaults to TorsoDataBehavior.FETCH_MODES.PULL if not set.
* __id or ids__ - duck-typed
  * __function()__ - expected to return the ids (either array or single primitive) to track with the private collection or to the single id to set on the singleton model.
  * __String[] or Number[]__ - the ids to use directly.
  * __String or Number__ - the id to use directly.
* __idProperty__ - identifies the property to get the ids from (on another object).
  * __view:propertyName__ - pulls from a viewState property of the containing view.
  * __model:propertyName__ - pulls from the model of the containing view.
[...]

PR #291, opened 2016-11-21
Data behavior

mandragorn:
Initial commit to get things started.


kentmw:
So, are you just setting up initialization/property setting?

mandragorn:
yes, I'll be pushing more commits to this branch as I go.

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,305 @@
...
+    },
+
+    /**
+     * Get the full data object contents (either an array of model attributes or a single model attribute based on the
+     * value of returnSingleResult) or the value of a specific property if a single result is expected.
+     * @method getData
+     * @param singleResultProperty {String} the property to get from the model (only valid if returnSingleResult is true).
+     * @return {Object} containing the full contents of either the collection or model.
+     */
+    getData: function(singleResultProperty) {
```
mandragorn: getDataToJSON() or toJSON()

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,305 @@
...
+    },
+
+    /**
+     * Adds the
+     * @method prepare
+     * @override
+     */
+    prepare: function() {
+      var behaviorContext = Behavior.prototype.prepare.apply(this);
+      _.extend(behaviorContext, this.privateCollection.toJSON());
```
mandragorn: update to use this.toJSON()

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,305 @@
...
+        if (_.isUndefined(this.__ids.context)) {
+          context = this.view;
+        } else if (_.isFunction(this.__ids.context)) {
+          context = this.__ids.context();
+        } else if (_.isString(this.__ids.context)) {
+          context = _.result(this, this.__ids.context);
+        } else if (_.isObject(this.__ids.context)) {
+          context = this.__ids.context;
+        } else if (_.isFunction(this.__ids.context)) {
+          context = this.__ids.context();
```
mandragorn: validate context type

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,305 @@
...
+        } else {
+          throw new Error('Data Behavior ids: Invalid context.  Not a string, object or function.');
+        }
+
+        var property = this.__ids.property;
+
+        var propertyParts = property.split('.');
```
mandragorn: make more readable.

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,305 @@
...
+        normalizedIds = normalizeIds(ids);
+        if (!_.isUndefined(normalizedIds)) {
+          idsDeferred.resolve(normalizedIds);
+        } else if (!_.isUndefined(ids) && _.isFunction(ids.promise)) {
+          idsDeferred = ids;
+        } else {
+          idsDeferred.resolve([]);
+        }
+      } else if (_.isString(this.__ids.property)) {
+        var context;
```
mandragorn: split out into more helper functions.

modules/behaviors/DataBehavior.js
mandragorn: done

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,305 @@
...
+        if (_.isUndefined(this.__ids.context)) {
+          context = this.view;
+        } else if (_.isFunction(this.__ids.context)) {
+          context = this.__ids.context();
+        } else if (_.isString(this.__ids.context)) {
+          context = _.result(this, this.__ids.context);
+        } else if (_.isObject(this.__ids.context)) {
+          context = this.__ids.context;
+        } else if (_.isFunction(this.__ids.context)) {
```
michaelpelletier: Is this just a duplication of 263? I'm also pretty sure that _.isObject will fire on a Function, so I'm not sure if this will even get hit?

kentmw: you'll want to check if it's a function first, as normal objects will report as an object but not a function whereas functions will report as both.

michaelpelletier: Yeah, for sure. But isn't this still a redundant check? Like, I think it works like this:

```
if (_.isUndefined(this._ids.context)) {
  context = this.view;
} else if (_.isFunction(this.__ids.context)) {
  // Fires on Function
  context = this.__ids.context();
} else if (_.isString(this.__ids.context)) {
  // Fires on String
  context = _.result(this, this.__ids.context);
} else if (_.isObject(this.__ids.context)) {
  // Fires on Object or Function, but Function has already been hit, so will only fire on Object.
  context = this.__ids.context;
} else if (_.isFunction(this.__ids.context)) {
  // Fires on Function, but Functions have already been hit, so is never called.
  context = this.__ids.context();
}
```

kentmw: ah yep, I just assumed you were referring to line 267 instead of 263. You are correct, it's the exact same line and in the wrong place this time.

mandragorn:
My goal with creating a separate context object that evaluates for the property is that I'm hoping to reuse it when creating the change listeners.  If I don't use it for that I may convert it into a utility class w/o state instead.

mandragorn:
just merged the changes that went into master for the 0.6.17 & 0.6.18 releases.

mandragorn:
should the DataBehavior re-trigger all events from the private collection on itself?

i.e. if the private collection of the data behavior triggers a 'change:someIds' event, should the DataBehavior instance also trigger the same event with the same arguments?

mandragorn:
2 things left:
* Code cleanup
* Doc cleanup

mandragorn:
docs done.  Will wait on peer reviews before doing final cleanup.

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,608 @@
...
+      }, this);
+    },
+
+    /**
+     * Removes existing event listeners.
+     * @method _undelegateEvents
+     * @private
+     */
+    _undelegateUpdateEvents: function() {
+      var updateEvents = this.__parseUpdateEvents();
```
kentmw: is there any time where parsing between delegate and undelegate would change?

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,608 @@
...
+      if (!_.isUndefined(this.ids.property)) {
+        this.stopListeningToIdsPropertyChangeEvent();
+        var idsPropertyNameAndContext = this.__parseIdsPropertyNameAndContext();
+        var idsContext = idsPropertyNameAndContext.idsContext;
+        var canListenToEvents = idsContext && _.isFunction(idsContext.on);
+        if (canListenToEvents) {
+          this.__currentContextWithListener = idsContext;
+          this.__currentContextEventName = 'change:' + idsPropertyNameAndContext.idsPropertyName;
+          this.__dependsOnBehavior = idsPropertyNameAndContext.isBehavior;
+          if (this.__dependsOnBehavior) {
```
kentmw: i'd love a running example as comments

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,608 @@
...
+        };
+      }
+    },
+
+    /**
+     * Validates that the cache property is valid and if not throws an error describing why its not valid.
+     * @method __normalizeAndValidateCache
+     * @private
+     */
+    __normalizeAndValidateCache: function() {
```
kentmw: isn't this just validate. not much normalizing

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,608 @@
...
+        var parsedContextDefinition = this.__parseIdsPropertyNameAndContext();
+        var propertyName = parsedContextDefinition.idsPropertyName;
+        var context = parsedContextDefinition.idsContext;
+
+        ids = context && context[propertyName];
+        var propertyOnContextIsUndefined = context && _.isUndefined(ids);
+        var contextHasGetMethod = context && _.isFunction(context.get);
+        var contextHasToJSONMethod = context && _.isFunction(context.toJSON);
+
+        if (propertyOnContextIsUndefined && parsedContextDefinition.isBehavior && contextHasToJSONMethod) {
```
kentmw: thorough

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,608 @@
...
+    },
+
+    /**
+     * Removes the listener added by listenToIdsPropertyChangeEvent().
+     * @method stopListeningToIdsPropertyChangeEvent
+     */
+    stopListeningToIdsPropertyChangeEvent: function() {
+      if (this.__currentContextWithListener) {
+        if (this.__dependsOnBehavior) {
+          this.stopListening(this.__currentContextWithListener, 'fetched', this.retrieve);
```
kentmw: help me understand why you need a special case for if it depends on a behavior...
Ahh, i think i understand your last question about if the behavior should push events, so you don't need this second block

modules/FormModel.js
```diff
@@ -86,7 +86,7 @@
      */
     constructor: function(attributes, options) {
       options = options || {};
-      this.__cache = {};
+      this.cache = {};
```
kentmw: why this change?

modules/behaviors/DataBehavior.js
```diff
@@ -0,0 +1,608 @@
...
+   *                           Will get the property before the first '.' from the view and if it is an object will try to use a
+   *                           .get('propertyName') on it and set a 'change:' listener on it.
+   *                           If it is a string/number or array of string/number, then it will use that as the ids.
+   *                           Triggering a 'change:context' event on the behavior will cause it to stop listing to the
+   *                           old context and start listening to the new one defined by this property.
+   *     - context {Torso.Cell|Backbone.Model|Function} - object (or a function that returns an object) that fires change
+   *                           events and has a .get('propertyName') function. It isn't required to fire events -
+   *                           the change event is only required if it needs to refetch when the id property value changes.
+   *     Examples:
+   *       - { property: '_patientId' }
```
kentmw: this is the good stuff!

test/karma/dataBehaviorSpec.js
```diff
@@ -0,0 +1,1681 @@
...
+      $.mockjax.clear(this.routes['/myModel/ids|post']);
+      $.mockjax(MOCKJAX_ROUTE_WITH_OTHER_IDS);
+
+      var defaultBehaviorConfiguration = getBasicBehaviorConfiguration();
+      defaultBehaviorConfiguration.ids = {
+        property: 'viewState.idFromView'
+      };
+      var defaultBehavior2Configuration = getBasicBehaviorConfiguration();
+      defaultBehavior2Configuration.ids = {
+        property: 'behaviors.dataBehavior.otherIds'
```
kentmw: it grabs "otherIds" from the models retrieved? I would have thought it was "behaviors.dataBehavior.[something].otherIds

mandragorn:
fixes #282

modules/behaviors/DataBehavior.js
mandragorn: yea data behaviors don't publish change events on themselves, but I think it might be useful?  If so then I can drop a lot of this stuff.  Instead of 'fetched' I could listen for 'change:context' and fire that along with 'fetched', then I could normalize this to a consistent block of code.

modules/behaviors/DataBehavior.js
mandragorn: not under normal use cases.  Only if you modify updateEvents pragmatically.

modules/behaviors/DataBehavior.js
mandragorn: yea was trying to keep naming consistent, but you are correct.

modules/FormModel.js
mandragorn: I made them top level properties so that you can set them via extension instead of just via behaviorOptions.

```
DataBehavior.extend({
  cache: someCacheCollection
});
```

test/karma/dataBehaviorSpec.js
mandragorn: yea, but what's the [something]?

I updated the data behavior to proxy all events from the private collection (which includes change events).  I think this is cleaner and is along the same lines with how toJSON() was updated.

mandragorn:
This is the first set of examples.  I want to add more to match the different configurations.

modules/FormModel.js
mandragorn: nvm, thought this was in a different file.

mandragorn:
Notes:
* Revert FormModel .__cache -> .cache change.
* Move examples at top of file to tonic dev to save on code.
* Move the following api to data namespace:
  * toJSON() - no arguments
  * get('propertyName')
  * Event API - proxy all private collection events to data.
  * privateCollection is property of data object of DataBehavior.
  * move fetched & fetched:ids to .data
  * move prepare to .data
* Update APIs to inlucde a .data to reference the events from a data behavior.

modules/ListView.js
```diff
@@ -22,7 +22,9 @@
       if (view.__delayedRenderTimeout) {
         clearTimeout(view.__delayedRenderTimeout);
         view.__delayedRenderTimeout = null;
-        view.render();
+        if (!view.isDisposed()) {
```
mandragorn: this was a weird bug that started showing up consistently in tests with this commit - the delayed render was being triggered after the view was disposed.

These seemed like reasonable protections (in ListView and View).  I'm guessing it was a race condition that we hit due to large numbers of tests being added.

mandragorn:
docs are up at https://runkit.com/torso/databehavior/0.0.5 - please review if you have time.

docs/js/files/modules_ListView.js.html
```diff
@@ -287,6 +295,11 @@ <h1 class="file-heading">File: modules/ListView.js</h1>
     __itemContext: null,
     __renderWait: 0,
     __delayedRender: null,
+    /**
+     * @property __delayedRenderTimeout
+     * @private
+     * @type {Number}
```
michaelpelletier: Why `Number` and not `Integer`?

mandragorn: as far as I know Integer is not a javascript type, there is just Number.  And yuidoc knows about built-in types so it would link {Number} correctly, but {Integer} not correctly.

mandragorn:
fixed #282

PR #293, opened 2016-11-28
Simultaneous fetch

### LINKED ISSUE #292: Promise prematurely returns when doing simultaneous pulls on private collections for the same cached model id
relevant line: https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/mixins/cacheMixin.js#L143

That line removes any ids that are currently in the process of being fetched.  The problem is that the returned promise will resolve when the new ids complete instead of when all requested ids are available.

This becomes more obvious when requesting the same ids by multiple views at the same time.
Ex:
1) view1 fetches model by id 1
2) view2 pulls model by id 1
If 2) happens while 1) is in progress, then the remaining ids are empty and it will return immediately and imply that the model has been fetched, when it hasn't yet.

mandragorn:
fixes #292


modules/mixins/cacheMixin.js
```diff
@@ -363,41 +379,64 @@
...
+            delete collection.idsFetching[fetchId];
+            _.each(requestedIds, function(requestedId) {
+              if (collection.idPromises) {
+                var existingPromisesForId = collection.idPromises[requestedId];
+                collection.idPromises[requestedId] = _.without(existingPromisesForId, fetchPromise);
+              }
+            });
+          });
+
+        if (!collection.idPromises) {
```
kentmw: shouldn't this be done during the construction of the collection?

modules/mixins/cacheMixin.js
```diff
@@ -363,41 +379,64 @@
...
+              }
+            });
         }, options)
-        .always(function() {
-          delete collection.idsFetching[fetchId];
-        });
+          .always(function() {
+            fetchComplete = true;
+            delete collection.idsFetching[fetchId];
+            _.each(requestedIds, function(requestedId) {
```
kentmw: what does this block do and why is it needed now?

modules/mixins/cacheMixin.js
```diff
@@ -363,41 +379,64 @@
...
+                var existingPromisesForId = collection.idPromises[requestedId];
+                collection.idPromises[requestedId] = _.without(existingPromisesForId, fetchPromise);
+              }
+            });
+          });
+
+        if (!collection.idPromises) {
+          collection.idPromises = {};
+        }
+        if (!fetchComplete) { // fixes the sync case (mainly during tests when mockjax is used).
```
kentmw: is this useful outside of test cases? It's a big chunk to just satisfy how mockjax is used.

modules/mixins/cacheMixin.js
```diff
@@ -139,9 +139,24 @@
           pull: function(options) {
             options = options || {};
             //find ids that we don't have in cache
```
kentmw: Can you update the comment to explain the "stitching" going on?

mandragorn: done

modules/mixins/cacheMixin.js
mandragorn: this happens once the promise is resolved, and removes the pending promise for that id.

note that an id may have mulitple promises (if fetch was called multiple times and then a pull with the same id).  We want to wait for all of them to complete, this tracks them separately and removes the in-flight promises once they are done.

modules/mixins/cacheMixin.js
mandragorn: good point...

modules/mixins/cacheMixin.js
mandragorn: just the if (!fetchComplete) { is for the test, the internals would be run every time otherwise.

modules/mixins/cacheMixin.js
```diff
@@ -359,45 +383,62 @@
...
+            delete collection.idsFetching[fetchId];
+            _.each(requestedIds, function(requestedId) {
+              if (collection.idPromises) {
+                var existingPromisesForId = collection.idPromises[requestedId];
+                collection.idPromises[requestedId] = _.without(existingPromisesForId, fetchPromise);
+              }
+            });
+          });
+
+        if (!fetchComplete) { // fixes the sync case (mainly during tests when mockjax is used).
```
kentmw: Add a comment above this if statement explaining that we are storing/cataloging this promise

PR #294, opened 2016-11-29
Simultaneous fetch - #292 - fixing how ids without any promises are handled.

mandragorn:


modules/mixins/cacheMixin.js
```diff
@@ -148,8 +148,7 @@
 
             //find ids that we don't have in cache and aren't already in the process of being fetched.
             var idsNotInCache = _.difference(this.getTrackedIds(), _.pluck(parentInstance.models, 'id'));
-            var allExistingPromisesForIds = _.pick(parentInstance.idPromises, idsNotInCache);
-            var idsWithPromises = _.mapObject(allExistingPromisesForIds, _.isEmpty);
+            var idsWithPromises = _.pick(parentInstance.idPromises, idsNotInCache);
```
mandragorn: removed the case where an id could have an empty array.  _.mapObject(allExistingPromisesForIds, _.isEmpty); didn't do what I expected it to.

modules/mixins/cacheMixin.js
```diff
@@ -429,7 +428,12 @@
             _.each(requestedIds, function(requestedId) {
               if (collection.idPromises) {
                 var existingPromisesForId = collection.idPromises[requestedId];
-                collection.idPromises[requestedId] = _.without(existingPromisesForId, fetchPromise);
+                var remainingPromisesForId = _.without(existingPromisesForId, fetchPromise);
+                if (_.isEmpty(remainingPromisesForId)) {
+                  delete collection.idPromises[requestedId];
```
mandragorn: if its empty just remove the reference entirely.  this is cleaner anyway.

kentmw:
+2

PR #302, opened 2017-02-01
Fixes and updates for DataBehavior first pass.

### LINKED ISSUE #300: add getModel() and getModels() to DataBehavior.data
For convenience it would be nice to get the actual models instead of just the data.

Currently you need to do: `dataBehavior.data.privateCollection.get(0)` to get the first model in the collection.  Would be nicer to do: `dataBehavior.data.getModel()` (assuming returnSingleResult === true) or `dataBehavior.data.getModels()` when returnSingleResult === false.

### LINKED ISSUE #301: DataBehavior's prepare() errors out when rendering a view.
the data behavior expects the prepare of Behavior to return an object, it does not (it uses _.noop which returns undefined).

https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/Behavior.js#L61
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L265-L268

mandragorn:
fixes #300
fixes #301


modules/behaviors/DataBehavior.js
```diff
@@ -263,7 +263,7 @@
      * @override
      */
     prepare: function() {
-      var behaviorContext = Behavior.prototype.prepare.apply(this);
+      var behaviorContext = Behavior.prototype.prepare.apply(this) || {};
```
mandragorn: this is the only thing to fix #301

PR #310, opened 2017-02-10
Fix DataBehavior.data extension of Events.

### LINKED ISSUE #309: Data Behavior .data field's event bus is shared across all Data Behaviors
basically all of the .data properties of every DataBehavior instance share the same _listenerId and are in fact the same Event object.

Extension of the Events object went horribly wrong....

mandragorn:
Fixes #309


kentmw:
strange, but reasonable.

Issue #327, opened 2017-05-22
FormModel define "models" attribute on the class level


kentmw:
You should be able to specify the models that are going to be tracked on the class level, not just the initialization options.

PR #333, opened 2017-06-27
 #327 - Define 'models' attibute on class level for FormModel

fyliu24:
Adding the ability to specify the models that are going to be tracked
 on the class level instead of just initialized options


docs/js/files/modules_FormModel.js.html
```diff
@@ -161,6 +161,12 @@ <h1 class="file-heading">File: modules/FormModel.js</h1>
     mapping: undefined,
 
     /**
+     * @property models
+     * @type Object
+     **/
+    models: undefined,
```
kentmw: What are our thoughts about existing methods relating to "models"? Like `getTrackedModel`, `getTrackedModels` or `setTrackedModel` or `setTrackedModels`.

I guess the question is two-fold: Will people be tempted in changing your new `this.models` directly hoping it calls these methods under the hood? And should methods like `setMapping` and `setMappings` default to use `this.models`?

fyliu24: I'm not sure...what about mapping? Does it do anything right now when people change this.mapping?

fyliu24: I feel like for now, this is just making it easier for people to define the models on class level when they are defining new form model that extends the FormModel. Not sure what's the use case when people will want to change the models after initialization?

kentmw: Eh, I guess you're right the mapping field follow the same behavior. Then the only question is should `setMapping` use this.models if they exist?

fyliu24: You mean when "models" is not given as an input to the setMapping method? That sounds a little confusing to me.

kentmw: Okay, it looks like the it falls back on existing tracked models when you update the mappings. The only thing you can't do is call `this.setMapping(...)` and hope it immediately copies the values. You'll have to do a `this.pull()` afterwards

docs/js/files/modules_FormModel.js.html
```diff
@@ -161,6 +161,12 @@ <h1 class="file-heading">File: modules/FormModel.js</h1>
     mapping: undefined,
 
     /**
+     * @property models
+     * @type Object
```
kentmw: can you add to this javadoc and the mapping javadoc to see the constructor for proper form/options for these fields?

docs/js/files/modules_FormModel.js.html
kentmw: I'm fine the way this is for now. Just lots of ways to setup a form model. It's a little concerning to make sure we've covered what the expected behavior is for all the ways. You can set mapping OR models OR both on the class level. At run time, you can set new mappings OR new mappings with new models OR just new models.

docs/js/files/modules_FormModel.js.html
fyliu24: sure

docs/js/files/modules_FormModel.js.html
```diff
@@ -155,12 +155,22 @@ <h1 class="file-heading">File: modules/FormModel.js</h1>
...
     /**
+     * Map from aliases (either model names or computed value names) to mappings.
+     * Please refer to the documentation on the constructor about the form and options for this field.
      * @property mapping
      * @type Object
      **/
     mapping: undefined,
 
     /**
+     * Map from model instances to model aliases.
```
kentmw: from model aliases to model instances

fyliu24: done

Issue #307, opened 2017-02-06
Track Loading state of DataBehavior


mandragorn:
Automatically track loading as part of the widget state so that views can render differently based on whether the data behavior is loading or completed.

Note: this may already be tracked in the collection state, just need to make it clear state in the behavior and easy to access.

PR #338, opened 2017-09-07
 #307 - Expose loading data loading info

mandragorn:
* Added 3 methods and exposed them on the DataBehavior directly and
  on the dataBehavior.data property.
** isLoading() - if anything is loading (ids or objects).
** isLoadingIds() - if ids are loading.
** isLoadingObjects() - if objects are loading from ids.


mandragorn:
@kentmw - another update to peer review.

modules/behaviors/DataBehavior.js
```diff
@@ -220,6 +220,8 @@
 
       Behavior.apply(this, arguments);
 
+      this.set('loadingIds', 0);
```
fyliu24: Any reason for using 0 instead of true/false?

mandragorn: yes for simultaneous requests.  If one finishes you don't want to say you are done if there is still another one in progress.

fyliu24: oh I see

mandragorn:
rebase

docs/modules/behaviors/DATA_BEHAVIOR.md
```diff
@@ -808,4 +824,23 @@ successEventPayload = {
   * "behaviors.behaviorAlias:[event name]" - arbitrary event triggered by another behavior on this view.
   * "behaviors.behaviorAlias.[property name]:[event name]" - arbitrary event triggered by another behavior's property on this view.
   * "behaviors.behaviorAlias.data:[event name]" - arbitrary event triggered by another behavior's data property on this view (specific example of `behaviorAlias.[property name]:[event name]`).
-  * { '[event name]': < object (or function returning an object) that the event is triggered on > } - arbitrary "event" triggered on the supplied object.
\ No newline at end of file
+<<<<<<< HEAD
```
kentmw: Merge conflicts

modules/behaviors/DataBehavior.js
```diff
@@ -774,6 +820,36 @@
     },
 
     /**
+     * Determine if behavior is loading ids or objects.
+     * @method isLoading
+     * @return {Boolean} true - the behavior is loading objects or ids.
+     *                   false - the behavior is not loading objects or ids.
+     */
+    isLoading: function() {
```
kentmw: just wondering how there's two isLoading methods in this file. It looks like they're different object, maybe a view and behavior? idk.

modules/mixins/loadingMixin.js
```diff
@@ -29,28 +29,30 @@
         base.call(this, args);
         this.loadedOnceDeferred = new $.Deferred();
         this.loadedOnce = false;
+        this.loadingCount = 0;
+        // Keeping loading boolean since it is technically public and may be in use.
```
kentmw: this is a comment that makes sense in the context of this patch but it kind of confusing later on. Maybe word it like:
"loading is a convenience flag that is the equivalent of loadingCount > 0"
You may also say it's deprecated if you'd like to remove it later, though I can see its use still

docs/modules/behaviors/DATA_BEHAVIOR.md
mandragorn: they seem to have been fixed along the way already.

modules/behaviors/DataBehavior.js
mandragorn: One is the Data object the other is the Behavior.  new (DataBehavior()).data is a full-fledged private object defined in DataBehavior.

modules/mixins/loadingMixin.js
mandragorn: see #341

Issue #335, opened 2017-07-17
DataBehavior getIds fail when ids is null


fyliu24:
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L506

__getIds function will fail if ids returns null when trying to grab ids.then

PR #342, opened 2017-09-11
 #335: fix npe when ids function returns null

mandragorn:
* __getIds() was checking for !_.isUndefined() which excludes null.
  Just checking ids truthy should be enough since it is also checking
  that ids.then is a function.  The ids var just needs to exist and
  the falsey values of ids will not have a .then function.


mandragorn:
another one for peer review @kentmw

Issue #308, opened 2017-02-06
DataBehavior trackids fails if it returns undefined.


mandragorn:
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L255

if ids is undefined there, then it should default to an empty array... possibly __getIds() should never return undefined.

kentmw:
@mandragorn is line 255 still correct?

mandragorn:
my guess is that it moved to https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L251

mandragorn:
line 251 is no longer correct either :(

PR #343, opened 2017-09-11
#308: process [] when ids deferred resolves null

mandragorn:
* __getIds() can return null if the deferred promise returned is
      resolved to null or undefined.  This makes sure that it always
      returns [] if null or undefined and the original value otherwise.


mandragorn:
@kentmw another if you are available.

Issue #336, opened 2017-07-17
DataBehavior shouldn't trigger a fetch if ids returns undefined and the collection currently empty


fyliu24:
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L508
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L524

When normalizedIds is undefined, it returns a promise with empty array of ids which will trigger a trackAndFetch of empty ids and then a fetchSuccess call which will trigger 'fetched'.
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L253
https://github.com/vecnatechnologies/backbone-torso/blob/master/modules/behaviors/DataBehavior.js#L258

When a view is initializing, the data behavior on it is triggered when the ids is usually still undefined. This shouldn't trigger a 'fetched' event. Since it's technically not fetching anything.

mandragorn:
@fyliu24 do you have an example of this that I can test against and the desired outcome?

PR #344, opened 2017-09-12
 #336 - DataBehavior update when to skip retrieval

mandragorn:
There are times when object retrieval isn't required and the
'fetched' event is undesirable.
* When the current ids is empty/undefined/null and the new ids is
  empty/undefined/null.
* When the function defining the ids returns something that indicates
  that ids shouldn't be updated and objects shouldn't be fetched.
  { skipObjectRetrieval: true }


mandragorn:
@kentmw - this was the last one I had queued up for DataBehavior and is the one that I'm least sure of.  Let me know what you think.

mandragorn:
2nd push is just a rebase after merging the other pull request.

mandragorn:
can you point out where I can make the code easier to read?

kentmw:
I tried. It's just one of those problems where what we want to do is complex and the code for databehavior itself is complex. So when we add more to it, it already starts off with a cloudy feeling. It's some awesome functionality, just the immense duck typing makes it hard to understand at first glance - i.e. there's twice as much duck typing, edge catching, and helper method code than core logic code.

My only thoughts on how to help is to refactor utility/helper functions out of this class. Also, I can't tell if making more general method name would help or hurt (i.e. `_skipRetrieveOnEmptyTrackedIdsAndNewIds` being `_skipRetrieveCheck`)

mandragorn:
ok, I'll merge for now and work on readability later.

PR #345, opened 2017-09-12
DataBehavior render when attached and active

mandragorn:
The view only needs to render if it is attached (unattached views
will render when they are attached) and is active (if the view isn't
active then it shouldn't be doing anything).


mandragorn:
@kentmw and another issue I ran across while doing a dry-run update of ps-frontend (one spot only adds the listener for rendering on the view's _activate method).

PR #349, opened 2017-09-15
DataBehavior triggers "fetched" when ids are empty

mandragorn:
* Even though fetching objects is skipped.
* This is different than the ids function returning
  skipObjectRetrieval: true - 'fetched' will still be skipped in this
  case unless forceFetchedEvent: true is passed along with
  skipObjectRetrieval: true.


modules/View.js
```diff
@@ -742,7 +743,7 @@
 
           if (prepareFieldValue && _.isFunction(prepareFieldValue.toJSON)) {
             prepareFieldsContext[prepareFieldName] = prepareFieldValue.toJSON();
-          } else if (!prepareFieldIsSimpleString && prepareFieldValueIsDefinedOnView) {
+          } else if (!prepareFieldIsSimpleString || prepareFieldValueIsDefinedOnView) {
```
mandragorn: random bug this logic was backwards and it wouldn't work if it was a string reference to a view field value.

Issue #353, opened 2017-10-02
Index of the item view in a list view accessible to the item view


kentmw:
It would be nice to have a method `getIndex()` on an item view or some way to know it's index. Optimally, I'd have access to it in the prepare method of the item view

Issue #252, opened 2016-04-26
Change api for overriding prepare()


pepperbc:
Currently when you override prepare(), you have to call the parent's prototype prepare(), which is difficult in the case of inheritance, and impossible in a mixin context. In general it's all a bit messy.

Proposal:

Change the signature of prepare() to prepare(context), where context is the present output of the default prepare method. The new default prepare would just be:

```
prepare(context) {
  return context;
}
```

Views could either overwrite or extend the passed in context.

kentmw:
I definitely agree that we should make overriding prepare easier. But, I'd rather not have an argument because I think it would confuse people that they should pass in paramenters to the prepare method. We made it not take arguments at the beginning because we went down a rabbit hole of preparing different results by passing in arguments. We decided that prepare method was always just a reflection of the current state, and that arguments weren't needed.

That said, because prepare is really a private function, we could have the render method invoke `_prepare` that invokes a null op `prepare` (the one you override) and then just adds to it the view/model defaults using `_.defaults` which would only write it if someone didn't already specify the view or model fields.

pepperbc:
Isn't prepare only called by torso internals, so that would be the only thing that would need to pass in the argument? The prepare method would still just be a reflection of the current state - the argument is just a way to do inheritance.

kentmw:
It's a private method but it's not a torso private method. If someone overrides _updateDOM, there have been a number of times where people choose between templates or reinvoke `this.templateRender(this.$el, this.template, this.prepare());` I don't want to make it hard for someone to invoke `this.prepare();`

mandragorn:
another option is we don't pass it in, but merge it for you.

if you define _prepare() then prepare() will return: 
`_.extend(View.prototype.prepare.apply(this), this._prepare())`

kentmw:
I like Josh's twist on this.

kentmw:
How about:
```
prepare: function() {
  var context = { ... };
  return this._prepare(context);
}

_prepare: function(context) {
  return context;
}
```

mandragorn:
:+1:

arikwex:
```
var BASE = Backbone.View.extend({
  constructor: function() {
    this.value = 123;
  },

  prepare: function() {
    return {
      baseContext: this.value
    }
  }
});

var CHILD = BASE.extend({
  constructor: function() {
    this.value = 456;
  },

  prepare: function() {
    var ctx = BASE.prototype.prepare.apply(this);
[...]

Issue #118, opened 2015-08-31
Add getter shortcut on view for viewState properties


kentmw:
Right now it's too bulcky to use the view's viewState. I have to do: this.viewState.get('foo'). At least add: this.get('foo') to alias the view state.

Also, should viewState, just be "state". Or maybe even just have cell logic mixed in to view.

kentmw:
Added `has` and `unset`

PR #354, opened 2017-10-24
#353, #252, #118 - view updates

kentmw:
added has and unset in view (for viewState) and added object mapping for prepareFields and _prepare option


modules/View.js
```diff
@@ -145,6 +161,16 @@
     },
 
     /**
+     * Augments the context with custom content.
+     * @param context {Object} the context you can modify
+     * @return {Object} [Optional] If you return an object, it will be merged with the context
+     * @method _prepare
+     */
+    _prepare: function(context) {
```
mandragorn: you could still use a _.noop here.

modules/View.js
```diff
@@ -748,7 +782,11 @@
           }
         }
       }
-      return prepareFieldsContext;
+      var context = this._prepare(prepareFieldsContext);
+      if (!_.isUndefined(context)) {
```
mandragorn: if _prepare() returns undefined then prepareFieldsContext is ignored.  I don't think that is what you want?

Please add tests for this so that we know it is working correctly for the corner cases and that it continues to work correctly :).

kentmw: do the tests run for you? it breaks when i run them

kentmw: also, good catch.

modules/View.js
kentmw: I know, but I thought that adding this empty function would be easier for people looking to see how _prepare works (i.e. there's a context argument

test/karma/viewPrepareSpec.js
```diff
@@ -0,0 +1,106 @@
...
+      });
+      var prepareView = new PrepareView();
+      var context = prepareView.prepare();
+      expect(context.foo).toBe(123);
+      expect(context.view).toBeDefined();
+      expect(context.view.bar).toBe(444);
+    });
+
+    it('can override existing context', function() {
+      var PrepareView = TorsoView.extend({
```
mandragorn: should this set bar in view state to make sure it is being overwritten?  A little bit ambiguous since you are overwriting something that doesn't have anything set on it.  Not a huge deal.

test/karma/viewPrepareSpec.js
```diff
@@ -0,0 +1,106 @@
...
+      });
+      var prepareView = new PrepareView();
+      var context = prepareView.prepare();
+      expect(context.foo).toBe(555);
+      expect(context.view).toBeDefined();
+      expect(context.view.bar).toBe(444);
+    });
+
+    it('can return an overriding context', function() {
+      var PrepareView = TorsoView.extend({
```
mandragorn: should this set bar in initialize?

test/karma/viewPrepareSpec.js
```diff
@@ -0,0 +1,106 @@
...
+        initialize: function() {
+          this.set('foo', 444);
+        },
+
+        _prepare: function(context) {
+          context.foo = 123;
```
mandragorn: should this be context.view.foo?  To match with view.foo set in initialize?  Or do you want to add an expect that the context.view.foo is 444?

test/karma/viewPrepareSpec.js
kentmw: I'm overridding the whole view object.

test/karma/viewPrepareSpec.js
kentmw: yea, that wasn't great

test/karma/viewPrepareSpec.js
kentmw: i need to show this better

PR #356, opened 2017-11-06
Adding viewMap and also untracking shared views during render

kentmw:


modules/View.js
```diff
@@ -31,6 +31,9 @@
       if (name === 'change' || name.indexOf('change:') === 0) {
         View.prototype.trigger.apply(this.view, arguments);
       }
+      if (name.indexOf('change:hide:') === 0) {
```
mandragorn: this makes me cringe for some reason, but I think its ok.

modules/View.js
```diff
@@ -52,6 +55,7 @@
     behaviors: null,
     templateRendererOptions: undefined,
     prepareFields: null,
+    viewMap: null,
```
mandragorn: alternative: views

modules/View.js
```diff
@@ -191,20 +195,25 @@
         return $.Deferred().resolve().promise();
       }
       this.__updateInjectionSiteMap();
-      this.trigger('render:before-dom-update');
+      this.trigger('render:before-detach-views');
       this.detachTrackedViews();
+      this.trigger('render:before-dom-update');
```
mandragorn: this moved, are there implications we need to worry about?

modules/View.js
```diff
@@ -696,6 +705,55 @@
     //************** Private methods **************//
 
     /**
+     * Attaches views using this.viewMap. The API for viewMap looks like:
+     * viewMap: {
```
mandragorn: what about if the injection point is a selector and not inject="" (I think torso supported using something other than inject attribute).

modules/View.js
```diff
@@ -696,6 +705,55 @@
...
+            config = config.call(this);
+          }
+          if (config instanceof Backbone.View) {
+            trackedView = config;
+          } else if (_.isObject(config)) {
+            options = config.options;
+            config = config.view;
+          }
+          if (!trackedView) {
+            if (_.isString(config)) {
```
mandragorn: maybe order this chain of ifs in most likely to least?  To prevent checking unlikely things.  Unless it already is, I didn't really think about it.

modules/View.js
kentmw: reverted

modules/View.js
kentmw: It's either the injection site element itself or the [inject='...'] attribute. There's no real good way to let them specify the injection site element in a map like this and I've never seen anyone do it that way anyway. I think this is okay

modules/View.js
kentmw: it's low risk, so maybe that makes seeing if the potential is worth it?

modules/View.js
kentmw: going with injectionSites

modules/View.js
kentmw: done

modules/View.js
mandragorn: I think the cringe was because it calls render directly.  I'm used to pushing render to be event triggered to prevent unnecessary renders.  But this is event triggering so... it was only a hesitation not a reason not to.

PR #360, opened 2017-11-14
 #359: Support objects with array syntax in Feedback

beiyi0207:
To be able to support object-backed array notation.


modules/View.js
```diff
@@ -1479,8 +1479,12 @@
      * @return {Array<String>} The fully expanded subattribute names
      */
     __generateSubAttributes: function(attr, model) {
-      var i, attrName, remainder, subAttrs, values,
-        firstBracket = attr.indexOf('[]');
+      var attrName;
```
kentmw: i feel like this took a backstep in readability

test/karma/viewFeedbackSpec.js
```diff
@@ -122,4 +122,88 @@ describe('A View', function() {
...
+      expect(this.objectFeedbackView.change).toBe(1);
+      expect(this.objectFeedbackView.indexMap.bar).toBe('abc');
+      this.objectFeedbackView.resetIndexMap();
+      this.objectFeedbackView.$el.find('[data-model="foo[abc]"]').change();
+      expect(this.objectFeedbackView.change).toBe(2);
+      expect(this.objectFeedbackView.indexMap.bar).toBe('abc');
+      this.objectFeedbackView.resetIndexMap();
+      this.objectFeedbackView.$el.find('[data-model="foo[def]"]').change();
+      expect(this.objectFeedbackView.change).toBe(3);
+      expect(this.objectFeedbackView.indexMap.bar).toBe('def');
```
kentmw: neat!

modules/View.js
beiyi0207: Use
```
var remainder, subAttrs, values, indexes,
    firstBracket = attr.indexOf('[]');
```
you mean?

kentmw: yea. admittedly, im very much in the mindset of es6 now that doesn't have to hoist. but, having them all on one line is my least favorite

Issue #362, opened 2017-12-08
Update View and Handlebar Helpers to support both arrays and objects


beiyi0207:

mandragorn:
@beiyi0207 - is this resolved?

beiyi0207:
Resolved.

PR #363, opened 2017-12-12
Torso #362: Update View and Handlebar Helpers

beiyi0207:
- Updated View.js and handlebarsUtils.js to support both array and object notations
- Updated comments in View.js and handlebarsUtils.js for better documentation
- Added Jasmine test for handlebarsUtils.js
- Added "noValueAttr" param for "formAttr" helper. When "noValueAttr" is set to true,
  "formAttr" will not generate a "value" attribute.


modules/handlebarsUtils.js
```diff
@@ -28,19 +28,30 @@
     /**
      * Usage: {{labelFor 'fieldName' value="suffix"}}
      * Generates: for="field-name-suffix"
+     * Usage: {{labelFor 'fieldName[x].sub' value="demo" x=123}}
+     * Generates: for="field-name-123_sub-demo"
```
kentmw: You have -123_ on this example but \_abc\_ on the next example. Are you using underscores, dashes, or periods?

test/karma/helpers/feedbackViews.js
```diff
@@ -205,7 +205,7 @@ var ObjectFeedbackView = TorsoFormView.extend({
     }
   ],
 
-  template: '<input name="foo[abc]" data-model="foo[abc]" data-feedback="foo[abc]" /><input name="foo[def]" data-model="foo[def]" data-feedback="foo[def]" />',
+  template: '<input name="foo[abc]" data-model="foo.abc" data-feedback="foo[abc]" /><input name="foo[def]" data-model="foo.def" data-feedback="foo[def]" />',
```
kentmw: it's a bummer that the attributes are written differently. the feedback can't do foo.abc, right?

modules/handlebarsUtils.js
beiyi0207: No... I was wondering about it. The current behavior is when the value is a number, it returns a dash before the value (e.g. field-name-123), but if the value is a string, it returns an underscore before the value instead (e.g. field-name_abc")

Should I figure out how to make both use dashes?

test/karma/helpers/feedbackViews.js
beiyi0207: I don't think it can...

modules/handlebarsUtils.js
```diff
@@ -126,22 +151,31 @@
     });
```
beiyi0207:
```
dotsRemoved = camelCaseRemoved.replace(/\./g, function() {
   return '_';
}
```

Is this what makes the string value use "_" instead of "-"?

kentmw:
My two concerns are: the fact that we have to use two different syntax's inside one field:
`data-model="foo.bar"` and `data-feedback="foo[bar]"`
and that the helper generates different prefixes depending on if the value is a number or string `-123` or `_foo`.

The second one I don't care too much about as it's just the handlebars and just pertains to the id and name attribute.

The first one is a bummer. If we merge this, we should create an issue to consolidate those.

beiyi0207:
I opened a new issue #364 to address the above concerns.

PR #365, opened 2018-01-19
Behavior dispose

mandragorn:
* Fixing security vulnerability
* Proving to myself that dispose is called.
* Updating dispose in behavior to follow the same strategy as dispose in the view for consistency.


mandragorn:
merged into https://github.com/vecnatechnologies/backbone-torso/pull/366

PR #366, opened 2018-01-19
Add a retrieve promise to DataBehavior

mandragorn:
* The goal is to have a promise that resolves/rejects immediately if
the behavior has already retrieved data or waits until the next fetch
completes before resolving or rejecting.  Resolving or rejecting is
based on the success/failure result of the fetch.

### DIFF
--- a/modules/Behavior.js
+++ b/modules/Behavior.js
@@ -18,7 +18,7 @@
     'before-detached-callback':  '_detached',
     'before-activate-callback': '_activate',
     'before-deactivate-callback': '_deactivate',
-    'before-dispose-callback': '_dispose',
+    'before-dispose-callback': 'dispose',
     'render:before-attach-tracked-views': 'attachTrackedViews',
     'render:begin': 'prerender',
     'render:complete': 'postrender',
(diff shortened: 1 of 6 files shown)


modules/Behavior.js
```diff
@@ -201,15 +201,34 @@
...
       this.stopListening();
       this.off();
-    }
 
+      this.__isDisposed = true;
+    },
+
+    /**
+     * Method to be invoked when dispose is called. By default calling dispose will remove the
+     * view's on's and listenTo's.
```
kentmw: behavior's on's and listenTo's

modules/behaviors/DataBehavior.js
```diff
@@ -860,10 +903,8 @@
      * @private
      */
     _dispose: function() {
+      Behavior.prototype._dispose.apply(this, arguments);
```
kentmw: There's no need to do the apply here if the parent is just a Torso.Behavior

modules/behaviors/DataBehavior.js
```diff
@@ -1036,6 +1077,7 @@
     dispose: function() {
```
kentmw: so you have a public dispose exposed on your behavior? the `this.off` and `this.listenTo` are already being performed by the hidden __dispose. Why not just move the `this.privateCollection.dispose()` to the _dispose you already have?

modules/Behavior.js
mandragorn: done

modules/behaviors/DataBehavior.js
mandragorn: done

modules/behaviors/DataBehavior.js
mandragorn: this is on the .data object, not on the DataBehavior itself.  The Data object starts on line 880.

Issue #374, opened 2018-02-21
collection remove method should also remove the id from tracked ids


fyliu24:
this.collection.remove(model) currently only remove the model from the collection but doesn't remove the id from tracked ids

PR #376, opened 2018-03-13
 #374: Remove model id from tracked ids on remove

mandragorn:
* Private collections will now remove the model's id from the tracked
ids when the model is removed from the collection using the .remove()
method.


modules/mixins/cacheMixin.js
```diff
@@ -185,6 +185,22 @@
...
+           * @param modelIdentifier {*} same duck-typing as Backbone.Collection.get():
+           *                              by id, cid, model object with id or cid properties,
+           *                              or an attributes object that is transformed through modelId
+           */
+          remove: function(modelIdentifier) {
+            var model = this.get(modelIdentifier);
+            parentClass.remove.apply(this, arguments);
+            if (model) {
+              var trackedIdsWithoutModel = this.getTrackedIds();
+              trackedIdsWithoutModel = _.without(trackedIdsWithoutModel, model.id);
```
kentmw: tracked ids are always model.id right? never by some other value

mandragorn: yes that is how the cache collection is setup?  Let me look closer.

mandragorn: addModelAndTrack() uses model.id, but you could technically use the model's cid (although you'd be sending that to the server which would be weird) or an object with an id property (or property matching the idAttribute of the model).  Note model.id is already normalized for idAttribute value (equivalent to model.get(model.idAttribute)).

PR #381, opened 2018-04-19
Fix #380 - misnamed Backbone reference

### LINKED ISSUE #380: FormView amd require of backbone uses incorrect casing
https://github.com/vecnatechnologies/backbone-torso/blob/534cf7e3c4ca3cb616a70e39c1df9104f3cd7b3c/modules/FormView.js#L4

Should be  define(['underscore', 'backbone',...
not define(['underscore', 'Backbone',

mikeobr:
Fix for #380


kentmw:
You need to rebuild the project so that the bundle is remade. `npm run build`

mikeobr:
Amended rebuild bundle files.

PR #385, opened 2018-04-25
Registry for Torso

kentmw:


modules/mixins/modelMixin.js
```diff
@@ -0,0 +1,34 @@
...
+   * @namespace Torso.Mixins
+   * @class  modelMixin
+   * @author kent.willis@vecna.com
+   */
+  return {
+    __postInitialize: function() {
+      registry.modelInitialized(this);
+    },
+
+    dispose: function() {
```
mandragorn: should we be worried about this keeping references to models that would have otherwise been un-referenced and therefore garbage collected?

modules/registry.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+
+    __initialize: function(obj, cache) {
+      cache[obj.cid] = obj;
+      this.listenTo(obj, 'before-dispose', function() {
+        delete cache[obj.cid];
+      });
+    },
+
+    disposeAll: function() {
```
mandragorn: I like all of the disposeAll methods.

Docs would be nice for the methods in the registry, though.

modules/registry.js
```diff
@@ -0,0 +1,83 @@
...
+    viewInitialized: function(view) {
+      this.__initialize(view, this.views);
+    },
+
+    serviceInitialized: function(service) {
+      this.__initialize(service, this.services);
+    },
+
+    __initialize: function(obj, cache) {
+      cache[obj.cid] = obj;
```
mandragorn: should we worry about objects w/o cids?  I guess that is all supplied by Backbone core so it is ok.

package-lock.json
```diff
@@ -0,0 +1,14396 @@
+{
```
mandragorn: is this a new npm thing?  Does it sync with yarn.lock?

modules/registry.js
mandragorn: I like all of the dispose functions.  These will make cleanup between tests easier.  There is still an issue around singletons being disposed by this and I'm getting some errors when I do similar things with views in my tests that I would like to track down before we release this.

Can we get docs for all of these methods?

And unit tests?

modules/mixins/modelMixin.js
mandragorn: should we be worried about keeping references to models that would normally have gotten garbage collected?  I.e. their dispose() isn't called as part of any of the lifecycles right now.

modules/ServiceCell.js
```diff
@@ -15,7 +15,11 @@
    * @class  ServiceCell
    * @author kent.willis@vecna.com
    */
-  var ServiceCell = Cell.extend({ });
+  var ServiceCell = Cell.extend({
+    __postInitialize: function() {
+      registry.serviceInitialized(this);
```
mandragorn: registry is not imported (found while testing this... definitely needs unit tests).

Issue #386, opened 2018-04-26
make the isAttached function null-proof


beiyi0207:
in `view.js`, the `isAttached` should also check whether `view.$el` is defined, so it is null-proof.

PR #387, opened 2018-04-26
#386 Make isAttached() in View.js null-proof

beiyi0207:


modules/View.js
```diff
@@ -405,7 +405,7 @@
      * @method isAttached
```
mandragorn: run the build so that you get all of the doc and bundle changes.

```
yarn build
```
or
```
npm run build
```

PR #395, opened 2019-05-29
Fix data model binding for numeric string key

fyliu24:
Javascript treats object key differently for numeric literals. It will always be coerced into a string but cannot be accessed by dot notation. Whereas, the bracket notation will work for this case. (See details in this post: https://mathiasbynens.be/notes/javascript-properties). Therefore, our data binding currently won't work for a key like '123'.

Update the data-model binding to use array notation instead of dot notation for numeric literals.


fyliu24:
All those failed checks seem unrelated to my changes...should I be worried about them?

PR #401, opened 2019-08-23
Fix snyk vulnerabilities

### LINKED ISSUE #400: Update torso to be jquery 3 compatible
Ideally torso should work with jquery 2 and jquery 3 so that the project using torso isn't restricted.

mandragorn:
fixes #400


mandragorn:
Tested to work with jQuery version 2.2.4 and jQuery version 3.4.1

